PR #88, opened 2022-08-29
[iOS] Not to use SwiftGenPlugin package to avoid high cpu usage

ry-itto:
related issues:

- https://forums.swift.org/t/spm-plugin-with-binary-target-high-cpu-usage/59535
- https://github.com/SwiftGen/SwiftGenPlugin/issues/2

According to these issues, artifact-bundle causes this problem, but it isn't caused by artifact-bundle in my environment. So I will use own plugin instead of plugin package experimentally.

### DIFF
--- a/app-ios/Plugins/SwiftGenPlugin/Plugin.swift
+++ b/app-ios/Plugins/SwiftGenPlugin/Plugin.swift
@@ -1,0 +1,30 @@
+import Foundation
+import PackagePlugin
+
+@main
+struct SwiftGenPlugins: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
+        let configurations: [Path] = [context.package.directory, target.directory]
+          .map { $0.appending("swiftgen.yml") }
+
+        return try configurations.map { configuration in
+            return Command.prebuildCommand(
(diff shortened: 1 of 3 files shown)


app-ios/Plugins/SwiftGenPlugin/Plugin.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+import PackagePlugin
+
+@main
+struct SwiftGenPlugins: BuildToolPlugin {
+    func createBuildCommands(context: PluginContext, target: Target) async throws -> [Command] {
+        let configurations: [Path] = [context.package.directory, target.directory]
+          .map { $0.appending("swiftgen.yml") }
```
pchmelar, line 8: Hey 👋 Thanks for this temporary workaround, it worked great in Xcode 13. However, in Xcode 14 we experience issues, because it is looking for the `swiftgen.yml` config file in both places and if one of them doesn't exist, the whole command fails. 

The solution is to add `.filter { FileManager.default.fileExists(atPath: $0.string) }`.
Thanks to @Carrione for this idea! 🚀

Carrione: @pchmelar NP :)

PR #194, opened 2022-09-05
Add missing spaces in README.md

jisungbin:
## Overview

In the Special Thanks part, there is a space between the session and the mention, but only the last API Server session does not contain a space between the mentions. Added missing whitespace for document readability and unifying style.

## Screenshot

Before | After
:--: | :--:
 |


takahirom:
👀

PR #198, opened 2022-09-05
Remove redundant ripples

Nthily:
## Overview (Required)
- Remove redundant ripples

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
Thank you for simple PR👍

PR #200, opened 2022-09-05
Remove redundant label in AndroidManifest

ogapants:
## Issue
- none

## Overview (Required)
The activity-label in AndroidManifest is redundant if it is the same as the application-label.
As far as I know, currently the activity-label is displayed only in "Recent Apps" on older OSes.
There was a Lint warning, so I just deleted it for now, but nothing changed.

## Links
- https://developer.android.com/guide/topics/manifest/activity-element

## Screenshot
Before | After
:--: | :--:
 | nothing changed


takahirom:
> ogapants

It's been a long time! Where did you go 😄 
Thank you for your PR anyway.

PR #201, opened 2022-09-05
Fix ktlint import layout and match it with Android Studio's default order

### LINKED ISSUE #77: Problem with import order differs between Android Studio and KtLint
Formatting imports in Android Studio with SessionsViewModel causes this problem. 
I believe it is probably a KtLint or Spotless configuration issue.

kafumi:
## Issue
- close #77 

## Overview (Required)
- 1f1c6bf4f406e158db2a1111764289ce3e773e20: Set EditorConfig property `ij_kotlin_imports_layout` for ktlint
- 9a5093652a62da0c96e7914be6702cee5c56f1b0: Apply Spotless again with the updated configuration 

## Details
- The default import order of ktlint is alphabetic order when `android` flag is supplied
  https://github.com/pinterest/ktlint/blob/0.45.2/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/ImportOrderingRule.kt#L40
  - It's different from Android Studio's default settings 👇
    CleanShot 2022-09-06 at 00 50 01@2x
- We can configure the import order of ktlint by setting `ij_kotlin_imports_layout` property of EditorConfig
  https://pinterest.github.io/ktlint/rules/configuration-ktlint/#import-layouts
- Spotless does not respect the `.editorconfig` settings but we can provide them by using  `editorConfigOverride()` on Gradle script
  https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#ktlint
- So, we can change the import layout settings of ktlint by passing `ij_kotlin_imports_layout` EditorConfig property via `editorConfigOverride()` on Gradle script

## Tests
- I've checked that no diff is generated any more when applying "Optimize Imports" to all files on Android Studio (except for some files in `build-logic`)

### DIFF
--- a/build-logic/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/SpotlessPlugin.kt
+++ b/build-logic/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/SpotlessPlugin.kt
@@ -15,9 +15,11 @@ class SpotlessPlugin : Plugin<Project> {
                 kotlin {
                     target("**/*.kt")
                     targetExclude("**/build/**/*.kt")
-                    ktlint(
-                        libs.findVersion("ktlint").get().toString()
-                    ).userData(mapOf("android" to "true"))
+                    ktlint(libs.findVersion("ktlint").get().toString())
+                        .userData(mapOf("android" to "true"))
+                        .editorConfigOverride(
+                            mapOf("ij_kotlin_imports_layout" to "*,java.**,javax.**,kotlin.**,^")
+                        )
                 }
                 format("kts") {
                     target("**/*.kts")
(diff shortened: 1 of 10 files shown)


takahirom:
👀

build-logic/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/SpotlessPlugin.kt
```diff
@@ -15,9 +15,11 @@ class SpotlessPlugin : Plugin<Project> {
                 kotlin {
                     target("**/*.kt")
                     targetExclude("**/build/**/*.kt")
-                    ktlint(
-                        libs.findVersion("ktlint").get().toString()
-                    ).userData(mapOf("android" to "true"))
+                    ktlint(libs.findVersion("ktlint").get().toString())
+                        .userData(mapOf("android" to "true"))
+                        .editorConfigOverride(
+                            mapOf("ij_kotlin_imports_layout" to "*,java.**,javax.**,kotlin.**,^")
```
takahirom, line 21: Thanks! I couldn't find the solution 🙏

takahirom (approved):
Thank you for saving everyone!

Issue #153, opened 2022-09-04
Scroll up when taps current tab


t-yusaku:
**Idea Description**
We can scroll on the timetable and sessions list.
So, we want to scroll up to top when we tap current tab. 

**Reference images and links**
![image](https://user-images.githubusercontent.com/58973616/188334450-5f0047b3-c2de-44e0-a589-92c8ee3c622a.png)

PR #206, opened 2022-09-05
Scroll up when taps current tab #153

adakoda:
- added a LazyListState to SessionList
 - added multiple LazyListState and pass them to SessionTopBar and SessionList
 - added jump to top on SessionsTopBar when user click selected tab

## Overview (Required)
- Scroll up when taps current tab

## Links
- https://github.com/DroidKaigi/conference-app-2022/issues/153

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
@@ -6,6 +6,8 @@ import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.lazy.LazyListState
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.shape.RoundedCornerShape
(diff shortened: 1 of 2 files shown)


feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -82,11 +84,13 @@ fun Sessions(
 ) {
     val scheduleState = uiModel.scheduleState
     val pagerState = rememberPagerState()
+    val listStates = DroidKaigi2022Day.values().map { rememberLazyListState() } .toList()
```
takahirom, line 87: This is very minor, but how about changing the name to sessionsListListStates? The reason is that it could be mistaken for Timetable's.

adakoda: You are right. I have addressed your point. [07ce675](https://github.com/DroidKaigi/conference-app-2022/pull/206/commits/07ce675915159c71fddb84c668097d9235eca274)

This commit also include bug fix (This feature should not work on TimeTable tab.).

takahirom (approved):
Thank you for tackling a somewhat complex problem 😄.
I noticed that there is a problem with state hoisting and the widget being displayed depending on the condition, which makes it complicated. However, I don't have a solution.
In the meantime, thanks!

PR #209, opened 2022-09-05
Show session start time for each session list item.

### LINKED ISSUE #208: show start time on the session list
**Idea Description**

The default timetable view has the obvious timeline on the left side and we can easily confirm the start time for each session. That however does not apply to the session list, so we cannot easily check what time the sessions will be. But I'm sure we would like to easily check them on site.

Current list

image

A proposed solution (needs design agreement)

image

atsushieno:
## Issue
- close #208

## Overview (Required)
- implement/fix #208 , show session start time on the list.

Even if the list is going to contain the start time as I suggested there, the design would be arguable. So far this involves fixed size `width=100.dp` which may violate any of the design. Also extra `Column { }` may violate current code design that I'm not aware of any.

## Screenshot

The PR brings is the suggested change screenshot at #208.

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionListItem.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionListItem.kt
@@ -54,9 +54,15 @@ fun SessionListItem(
             )
 
             Row {
+                Column(modifier = Modifier.width(100.dp)) {
+                    KaigiTag(
+                        text = roomName.enTitle,
+                        backgroundColor = roomColor,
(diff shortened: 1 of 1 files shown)


takahirom:
Thank you! I think we need this feature. But we have a design for this.
https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=448%3A2009

Can you make an issue that applies the start time design and remove this start time tag?
image

takahirom:
Sorry, And we have a PR for it. I should apologize because I didn't create that issue 🙇  
https://github.com/DroidKaigi/conference-app-2022/pull/213

PR #213, opened 2022-09-05
Improve the session list ui

Nthily:
## Issue
-  #208

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionList.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionList.kt
@@ -1,17 +1,12 @@
 package io.github.droidkaigi.confsched2022.feature.sessions
 
-import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListState
-import androidx.compose.foundation.lazy.items
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.key
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.unit.dp
-import io.github.droidkaigi.confsched2022.model.Timetable
-import io.github.droidkaigi.confsched2022.model.TimetableItem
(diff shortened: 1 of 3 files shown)


feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionList.kt
```diff
@@ -1,30 +1,32 @@
...
-        modifier = modifier.fillMaxSize(),
-        verticalArrangement = Arrangement.spacedBy(32.dp)
+        modifier = modifier.fillMaxSize()
     ) {
-        items(timetable.contents, key = { it.timetableItem.id.value }) {
-            content(
-                it.timetableItem,
-                it.isFavorited
-            )
+        timetable.contents[0].timetableItem.startsAt
```
takahirom, line 21: Do we need this? 👀

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -182,13 +191,60 @@ fun SessionsList(
...
+            val endTime = LocalTime(
+                hour = endLocalDateTime.hour,
+                minute = endLocalDateTime.minute
+            ).toString()
+            var displayTime = true
+            if (index == 0) currentStartTime = startTime
+            else if (index > 0 && startTime == currentStartTime) {
+                displayTime = false
+            } else {
+                currentStartTime = startTime
```
takahirom, line 213: I'm concerned that this can cause unexpected behavior because I think the lambda isn't always called sequentially.

How about making the list in advance? 🩹 

```kotlin
val timeHeaderAndTimetableItems: List<Pair<String?, TimetableItemWithFavorite>> = remember(timetable) {
  ... 
}
SessionList(timeHeaderAndTimetableItems) { (timeHeader, timetableItemWithFavorite), index ->
```

takahirom (approved):
Thank you for your second PR!
Please fix the conflict 👍

takahirom:
Please fix conflict 🙏

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionList.kt
Nthily: no, I forgot to delete it 👀

takahirom (approved):
LGTM! Thank you for providing the schedule! ⏲️

PR #216, opened 2022-09-05
Improve timetable scroll accessibility

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- With this implementation, can hear how much the time table is being scrolled with Talkback.
  - Maybe it's better to display a list instead of time table if Talkback is enabled.🤔 
    - Like this → https://prtimes.jp/main/html/rd/p/000000016.000046712.html

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
@@ -3,6 +3,7 @@ package io.github.droidkaigi.confsched2022.feature.sessions
 import androidx.compose.animation.core.Animatable
 import androidx.compose.animation.core.exponentialDecay
 import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.focusGroup
 import androidx.compose.foundation.gestures.awaitFirstDown
 import androidx.compose.foundation.gestures.awaitTouchSlopOrCancellation
 import androidx.compose.foundation.gestures.drag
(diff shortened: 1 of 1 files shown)


feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
```diff
@@ -124,6 +131,28 @@ fun Timetable(
                         }
                     }
                 )
+            }
+            .semantics {
+                horizontalScrollAxisRange = ScrollAxisRange(
```
NUmeroAndDev, line 136: https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/ScrollAxisRange

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
```diff
@@ -124,6 +131,28 @@ fun Timetable(
...
+                    value = { -scrollState.scrollX },
+                    maxValue = { -scrollState.maxX },
+                )
+                verticalScrollAxisRange = ScrollAxisRange(
+                    value = { -scrollState.scrollY },
+                    maxValue = { -scrollState.maxY },
+                )
+                scrollBy(
```
NUmeroAndDev, line 144: https://developer.android.com/reference/kotlin/androidx/compose/ui/semantics/package-summary#(androidx.compose.ui.semantics.SemanticsPropertyReceiver).scrollBy(kotlin.String,kotlin.Function2)

takahirom:
Do you mean you can scroll the timetable by TalkBack? 📜

NUmeroAndDev:
>Do you mean we can scroll the timetable by TalkBack?

It's not.

You can scroll time table without this implementation even with Talkback enabled.
But Talkback doesn't tell you how far you have scrolled or how much more you can scroll.

By the way, scroll the content with two fingers in Talkback.

takahirom (approved):
Let's enhance usability for all users 👍

PR #225, opened 2022-09-06
add minifyEnabled for release

### LINKED ISSUE #138: Enable Proguard and see if it can run.
**Idea Description**
The title says everything.

Please add minifyEnabled = true for release and see if developRelease can run. The goal is that you launch the app. 🚀 
https://github.com/DroidKaigi/conference-app-2022/blob/main/app-android/build.gradle.kts

Nthily:
## Issue
- close #138 

## Overview (Required)
- Based on the size of the generated apk, it appears that proguard and minify are enabled 🤔
- image

## Screenshot
Before | After
 |

### DIFF
--- a/app-android/build.gradle.kts
+++ b/app-android/build.gradle.kts
@@ -33,6 +33,13 @@ android {
         }
     }
     buildTypes {
+        release {
+            isMinifyEnabled = true
+            proguardFiles(
+                getDefaultProguardFile("proguard-android-optimize.txt"),
+                "proguard-rules.pro"
+            )
+        }
         debug {
             signingConfig = null
         }


takahirom:
Can you launch the app?

Nthily:
yes, but I can't seem to load the timetable

takahirom:
Thanks! Can you make an issue that adds proguard rule?

takahirom:
Sorry let me think it over 🙇

takahirom:
I think like this.
`We can release with proguard > We can release without proguard > We can't release`
So, If you can add proguard rules. I can merge this.
I should apologize that I created a vague issue. If you would like to do that. Please do it 🙏

takahirom:
@Nthily 
Would you like to implement this PR? If not, let me close and can I create a new issue for it? 🙇

Nthily:
@takahirom OK, I found that even if I don't add proguard rules and minifyEnabled, it still can't show the sessions list, but I haven't tested the latest code, so let's close this PR first.

PR #215, opened 2022-09-05
:recycle: Update About screen

Corvus400:
## Issue
- #181
  - The CLOSE notation is removed because the task is not completely completed.
  - I have issued a PR, although it is on the way so as not to block the work of other implementers.

### What I did.

- Add and display image resources.
- Add and display string resources.
- Add fonts.
- Add Divider under external services such as Twitter.

### Things that have not been done yet.

- Adjustment of font display.
  - The fonts on the design and the implemented fonts are displayed differently.
- The title is not displayed in the AppBar.

## Links
- https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=400%3A2957

## Screenshot
Before | After
:--: | :--:
 |

PR #226, opened 2022-09-06
Add Auxiliary Information Section for About Screen

### LINKED ISSUE #181: Apply about screen design
**Idea Description**
please apply cool design for about screen following figma instruction.

[Figma](https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/Conference-App?node-id=0%3A1)

kittinunf:
## Issue
- close #181 with a row for others to connect with a click listener 

## Overview (Required)

This is a continuation of work on top of #215 as a foundation. I have added the extra info section. This should allow others who implement the connection for each row item to continue with their works in sub issue of #215

## Screenshot
Before | After
:--: | :--:
 | 

This is in 日本語

Screen Shot 2022-09-06 at 19 56 21


feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
```diff
@@ -129,9 +140,38 @@ fun About(
                     ),
                 color = Color(KaigiColors.neutralVariantKeyColor50)
             )
-            Column {
-                // TODO: Implementation of this part
-                // TODO: https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=421%3A1883
+            Column(modifier = Modifier.padding(horizontal = 17.dp)) {
```
takahirom, line 143: Probably this is 16dp 🙏

takahirom:
Almost perfect!

feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
kittinunf: Ah, sure! Done 💪

feature-about/build.gradle.kts
```diff
@@ -10,18 +10,20 @@ dependencies {
     implementation(projects.coreModel)
     testImplementation(projects.coreTesting)
 
-    implementation(libs.kermit)
-    implementation(libs.hiltNavigationCompose)
+    implementation(libs.accompanistPager)
```
kittinunf, line 13: I sort this alphabetically so it is easier to understand which one is which 😉

feature-about/src/main/res/values-en/strings.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<resources>
+    <string name="about_title">What is DroidKaigi?</string>
+    <string name="about_description">DroidKaigiはエンジニアが主役のAndroidカンファレンスです。\nAndroid技術情報の共有とコミュニケーションを目的に、2022年10月5日(水)〜7日(金)の3日間開催します。</string>
```
kittinunf, line 4: This probably needs a better translator so I leave them as 日本語 first.

feature-about/build.gradle.kts
takahirom: I thought sorting dependencies is not so good idea. But I think it is better than I thought 👍 
May I ask how you are sorting them?

takahirom (approved):
LTGM 👍

feature-about/build.gradle.kts
kittinunf: I use IdeaVim to sort this 😄 

https://vim.fandom.com/wiki/Sort_lines

PR #247, opened 2022-09-07
[skip ci] :wrench: Fix Fixed the part of the template that remains the same.

Corvus400:
## Issue
- Nothing

## Overview (Required)
- As the title says.
  - I followed last year's section and modified it.

## Links
- https://github.com/DroidKaigi/conference-app-2021/blob/main/LICENSE#L189

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Thanks 🙏 
For your reference, this is why it says DroidKaigi committee.
https://github.com/DroidKaigi/conference-app-2020/pull/292#discussion_r367420517

PR #263, opened 2022-09-07
Added Timetable room header

### LINKED ISSUE #205: Timetable room header
**Idea Description**
The timetable code is here
https://github.com/DroidKaigi/conference-app-2022/blob/main/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
image

surajsau:
## Issue
- close #205 

## Overview (Required)
- added Room headers

## Screenshot

https://user-images.githubusercontent.com/5493586/188887189-87fdd2de-f476-4415-b8cc-13cec53b4994.mp4

## Discussion
- While the design didn't specify for static headers, I implemented room headers as static considering that it'll be an useful context when scrolling through the sessions. Does it conform with the design?
- At present I've duplicated `timeTableColumnWidth = 192.dp` from `TimeTable.kt` as `roomHeaderWidth = 192.dp` separately in `Rooms.kt`. Would you suggest using a common value instead?


takahirom:
👀

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -154,19 +154,28 @@ fun Timetable(
...
-            ) { timetableItem, isFavorited ->
-                TimetableItem(
-                    timetableItem = timetableItem,
-                    isFavorited = isFavorited,
-                    modifier = Modifier
-                        .clickable(
-                            onClick = { onTimetableClick(timetableItem.id) }
-                        ),
-                )
+            Column(modifier = Modifier.weight(1f)) {
```
takahirom, line 157: 🆒

takahirom (approved):
LGTM 💯

takahirom:
Sorry I didn't see the discussion.

> While the design didn't specify for static headers, I implemented room headers as static considering that it'll be an useful context when scrolling through the sessions. Does it conform with the design?

I think it is OK.

> At present I've duplicated timeTableColumnWidth = 192.dp from TimeTable.kt as roomHeaderWidth = 192.dp separately in Rooms.kt. Would you suggest using a common value instead?

We should make it a common value but let's merge anyway!

PR #265, opened 2022-09-07
Set weight to the height of the text in TimetableItem

### LINKED ISSUE #91: Adjusts the maximum number of rows by the height of the timetable item
The size of TimeTable Item can change depending on the duration of the talks.
Therefore, we would like to adjust the maximum number of rows by their height.

yamacraft:
## Issue
- close #91 

## Overview (Required)
- weight of TimetableItem-Text and remove of maxlines setting.
- maxLines setting does not work well for changed the device font size.
  - The maxline of the text is automatically varied to match the display area of the Text.

## Screenshot
Before | After
:--: | :--:
 | 
 |

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt
@@ -24,8 +24,7 @@ import io.github.droidkaigi.confsched2022.model.TimetableItem
 fun TimetableItem(
     timetableItem: TimetableItem,
     isFavorited: Boolean,
-    modifier: Modifier = Modifier,
-    maxTitleLines: Int = 4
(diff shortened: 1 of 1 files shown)


takahirom:
👀

takahirom (approved):
I thought we have to adjust the maxTitleLines number. But what we have to do is just this! Very cool 🆒

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt
```diff
@@ -47,9 +46,9 @@ fun TimetableItem(
             color = Color.White,
             modifier = Modifier
                 .padding(bottom = 8.dp)
+                .weight(weight = 1f, fill = false)
```
takahirom, line 49: I didn't know the fill = false. So let me put an awesome label on this issue. 🏷️

PR #282, opened 2022-09-08
Make common value for timetable layout size

### LINKED ISSUE #266: Make a common value for layout size
**Idea Description**
There are values duplicated like timeTableColumnWidth = 192.dp and TimeTable.kt as roomHeaderWidth = 192.dp.
Can we make an object like this and make a common value?
Say, TimetableDefaults
https://github.com/android/nowinandroid/blob/900450eb059ce20d233ee611b4fd7fbaa0aa040e/core-designsystem/src/main/java/com/google/samples/apps/nowinandroid/core/designsystem/component/ToggleButton.kt#L102

wildstarzw:
## Issue
- close #266 

## Overview (Required)
- Add timetable layout size object and share between timetable layout and rooms layout
- About the naming of the size object, I simply use xxxSizes but please suggest something better if has 😛
- Is it better to always have the size object instead of private vals for consistency? 🤔


takahirom:
👀

takahirom (approved):
I think it is a good name for the current situation. Thanks!

PR #287, opened 2022-09-08
[Proposal] Simplify uiModel with reusable UiLoadState 

kittinunf:
## Issue

- N/A as this is a foundation work.

## Overview (Required)

The way that we create `UiModel` is nice and great. However, it can be much improved.

### Problem

Typical `UiModel` for VM in the current setup requires the following;

```kotlin
data class HogeUiModel(
  val hogeState: HogeState
) {
  sealed interface HogeState {
     data class Loaded

     object Loading
 
     companion object {
            fun of(...)
     }
  }
}
```

- The sealed interface is quite boilerplate-y and we need to be re-defined on all screens and/or `uiModel`.
- Usage of Result <-> xxxState is kinda redundant.

While it is nice, I think we can create a reusable construct that could simplify the setup and could be used in all of the possible `UiModel` in the app.

With the new setup, we introduce the `LoadState`. (Right now it lies in the `Result` but I think the usage of `Result` seems to be redundant after the introduction of `LoadState`)

Now the `UiModel` can become quite simpler and arguably easier.

(Use the `ContributorsScreen` with `ContributorsUiModel` as an example)

```kotlin

data class ContributorsUiModel(
    val contributorsState: LoadState<PersistentList<Contributor>>,
)
```

We define the Ui model as usual, however, it is merely a wrapper over the `LoadState`

```kotlin
val uiModel = moleculeScope.moleculeComposeState(clock = ContextClock) {
  //...

  ContributorsUiModel(contributorsResult)
}
```

Also, I don't think the `derivedOf` part is actually needed anymore.

The UI side of things will be clearer and easier to break into smaller components.

```kotlin
/// in composable function

when (uiModel.state) {
  is Error -> TODO() // error UI 
  LoadState.Loading -> { CircularProgressIndicator() } //loading UI
  is Success -> {
     // success UI
  }
} 
```

### Caveat

In this PR, I have only refactored the contributor screen to show as an example. If this proposal is accepted, I will move on to other screens.

## Links
-

## Screenshot
Nothing changes


takahirom:
Thank you for your proposal! Your insight will improve our application! Let me think it over 👀

kittinunf:
Sure. Let me know if you have questions or anything. I m keen to help out as much as I can and improve this codebase further ✨🌟⭐

core-ui/src/main/java/io/github/droidkaigi/confsched2022/ui/Result.kt
```diff
@@ -13,10 +13,41 @@ sealed interface Result<out T> {
...
 fun <T> Flow<T>.asResult(): Flow<Result<T>> {
-    return this
-        .map<T, Result<T>> {
-            Result.Success(it)
+    return this.map<T, Result<T>> {
+        Result.Success(it)
+    }.onStart { emit(Result.Loading) }.catch { emit(Result.Error(it)) }
+}
+
+sealed interface LoadState<out T> {
```
takahirom, line 21: I can understand what you want to do. But this model has the same data.
So probably we may be confused about which to use. So how about renaming it to UiLoadState.

takahirom:
Sorry to keep you waiting 🙇

core-ui/src/main/java/io/github/droidkaigi/confsched2022/ui/Result.kt
kittinunf: Sure I can rename this!

kittinunf: Done 🎇

takahirom (approved):
Looks great! Thank you for suggesting architectural change!

takahirom:
Can you take a look at the conflict? 🙏

kittinunf:
Fixed!

PR #289, opened 2022-09-08
Fix extrernal service image button to match Material Design Accessibility

hiraike32:
## Issue
none

## Overview (Required)
- On the About screen, external service image buttons are too small and difficult to tap. Material Design recommends that the size of buttons should be 48dp or more. So, I fixed buttons paddings and other layouts paddings.

## Links
- [Accessibility - Material Design](https://material.io/design/usability/accessibility.html#layout-and-typography)

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
+++ b/feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
@@ -108,7 +108,7 @@ fun About(
                     .padding(
                         start = 16.dp,
                         end = 16.dp,
-                        bottom = 32.dp,
+                        bottom = 14.dp,
                     ),
                 verticalArrangement = Arrangement.spacedBy(24.dp)
             ) {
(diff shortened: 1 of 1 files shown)


feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
```diff
@@ -255,13 +252,14 @@ private fun ExternalServiceImage(
...
-            .size(24.dp)
             .clickable {
                 navigateToExternalServices(
                     context = context,
                     serviceType = serviceType
                 )
-            },
+            }
+            .padding(12.dp)
+            .size(24.dp),
```
hiraike32, line 262: This changes make buttons a little wider than the Figma design. However, I think that the ease of pressing the buttons is more important than the design, so I changed it.

takahirom:
👀

feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
takahirom: Agree. I told it to the designer. And I'm waiting the response🙏

takahirom: The designer says it is OK! Thank you for telling us!

PR #302, opened 2022-09-08
Not to reuse Modifier instances more than necessary in SessionDetail.kt

### LINKED ISSUE #254: Not to reuse Modifier instances more than necessary in SessionDetail.kt
**Idea Description**
- I want to avoid passing the argument modifier of the composable fun to the child composable in the [SessionDetail.kt](https://github.com/DroidKaigi/conference-app-2022/blob/55948f565047505abb503899a77e40c2766f3ed5/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt) file.

- I want it like this↓.

```Kotlin
@Composable
fun SessionDetailSessionInfo(
    ...
    modifier: Modifier = Modifier,
) {
    Column(modifier = modifier) {  // pass modifier arguments to root composable
        Text(
            modifier = Modifier,  // not use modifier arguments
            ...
        )
        ...
   }
}
```

SSong-develop:
## Issue
- close #254 

## Overview (Required)
- avoid passing the argument modifier of the composable fun to the child composable 

## Links
- Nothing

## Screenshot
Nothing (I just change use of the composable function params, this can be checking code differ. so do not capture code snippet)

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt
@@ -316,14 +316,14 @@ fun SessionDetailSessionInfo(
     language: String,
     levels: PersistentList<String>,
 ) {
-    Column {
+    Column(modifier = modifier) {
         Text(
-            modifier = modifier,
+            modifier = Modifier,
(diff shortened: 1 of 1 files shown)


RyuNen344:
👀

takahirom (approved):
We definitely need this! Thanks!

PR #304, opened 2022-09-08
:sparkles: Add Added a simple test.

### LINKED ISSUE #300: Add easy test to announcement
**Idea Description**

We would like to create a test like AboutScreenTest. The test is to check the announcement screens title exists.

https://github.com/DroidKaigi/conference-app-2022/blob/main/feature-about/src/test/java/io/github/droidkaigi/confsched2022/feature/about/AboutScreenTest.kt

Corvus400:
## Issue
- close #300 

## Overview (Required)
- Test that the title of the notification screen is displayed.
  - I thought it would be better to add a test for each of JP and EN, but I couldn't figure out how to test for each language. 🤔

## Links
- Nothing.

### DIFF
--- a/feature-announcement/src/test/java/io/github/droidkaigi/confsched2022/feature/announcement/AnnouncementScreenTest.kt
+++ b/feature-announcement/src/test/java/io/github/droidkaigi/confsched2022/feature/announcement/AnnouncementScreenTest.kt
@@ -1,0 +1,23 @@
+package io.github.droidkaigi.confsched2022.feature.announcement
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import dagger.hilt.android.testing.HiltAndroidTest
+import io.github.droidkaigi.confsched2022.testing.RobotTestRule
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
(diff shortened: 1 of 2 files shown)


RyuNen344:
👀

feature-announcement/src/test/java/io/github/droidkaigi/confsched2022/feature/announcement/AnnouncementScreenTest.kt
```diff
@@ -0,0 +1,23 @@
...
+@RunWith(AndroidJUnit4::class)
+@HiltAndroidTest
+class AnnouncementScreenTest {
+    @get:Rule val robotTestRule = RobotTestRule(this)
+    @Inject lateinit var announcementScreenRobot: AnnouncementScreenRobot
+
+    @Test
+    fun checkAnnouncementVisible() {
+        announcementScreenRobot(robotTestRule) {
+            checkAnnouncementVisible()
```
takahirom, line 20: How about renaming checkAnnouncementVisible -> checkAnnouncementTitleVisible
Because it is not a little clear.

takahirom (approved):
Thanks!

PR #332, opened 2022-09-09
Update Zipline version from 0.91 to 0.94

### LINKED ISSUE #249: Update Zipline
**Idea Description**
Run zipline locally and see if it works when update.

https://github.com/DroidKaigi/conference-app-2022/pull/58

horie1024:
## Issue
- close #249

## Overview (Required)
- `ManifestVerifier.Companion.NO_SIGNATURE_CHECKS` to be passed when initializing ZiplineLoader.
- Context was deleted because it was no longer needed during initialization of ZiplineLoader.
- The load method of ziplineLoader now accepts serializersModule, so the parameter name is explicitly specified.

I tried to get zipline to work in my local environment, but it did not work. The command used to check was the following, which was checked by starting up a local server.

`./gradlew jsBrowserDevRun --info --stacktrace` 

The result was the same before and after the version update. So I have opened this pull request regardless of whether zipline works or not.

## Links
N/A

## Screenshot

N/A


takahirom (commented):
Looks good!

PR #336, opened 2022-09-09
:sparkles: Add Prevent classes not used in Jetpack Compose from appearing in suggestions.

Corvus400:
## Issue
- Nothing.

## Overview (Required)
- As the title says.

## Links
- https://github.com/GetStream/whatsApp-clone-compose/blob/main/.idea/codeInsightSettings.xml

## Movie

https://user-images.githubusercontent.com/13657682/189455030-745343c9-cc0a-404e-8739-c73516a125ec.mov

https://user-images.githubusercontent.com/13657682/189455034-df8f4783-4c80-455b-a1a4-ed1416703c54.mov

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,8 @@ local.properties
 *.iml
 .idea/*
 !.idea/copyright
+!.idea/codeInsightSettings.xml
+
 # Keep the code styles.
 !/.idea/codeStyles
 /.idea/codeStyles/*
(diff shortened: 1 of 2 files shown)


takahirom:
Thanks! I have a question about this. Can we use the excluded classes even if you exclude them? 👀 
Our application may be OK with these settings. But if we were a hybrid app, could it be troublesome?

takahirom (approved):
Thanks! 🔡

Corvus400:
@takahirom 
I believe it is only that it will not appear as a candidate for autocomplete.
It should not cause any problems such as not being able to compile.
I have confirmed that it can be used in the excluded classes by adding import yourself.😃👍

PR #337, opened 2022-09-10
Hoist URL handling to KaigiApp

### LINKED ISSUE #296: Hoist URL handling to KaigiApp
**Idea Description**

There are a lot of startActivity intent on screens to open web pages.
How about hosting the event and handling it in new `KaigiExternalNavigation`?

And we can also add navigateGitHub(xxx) for adding the special handling for the intent.

```
    kaigiAppScaffoldState: KaigiAppScaffoldState = rememberKaigiAppScaffoldState(),
    kaigiExternalNavigationController = rememberKaigiExternalNavigationController()
) {
aboutNavGraph(
...
  onLinkClick = kaigiExternalNavigationController::navigate
)
```

fornewid:
## Issue
- close #296

## Overview (Required)
- Hoist URL handling to KaigiApp
- Remove all `androidx.browser` dependencies in feature modules.

### DIFF
--- a/feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
+++ b/feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
@@ -1,11 +1,7 @@
 package io.github.droidkaigi.confsched2022.feature.about
 
-import android.content.ActivityNotFoundException
 import android.content.Context
-import android.content.Intent
-import android.net.Uri
 import androidx.annotation.StringRes
-import androidx.browser.customtabs.CustomTabsIntent
 import androidx.compose.foundation.Image
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.Arrangement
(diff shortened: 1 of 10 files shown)


feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
```diff
@@ -53,15 +49,17 @@ fun AboutScreenRoot(
     modifier: Modifier = Modifier,
     showNavigationIcon: Boolean = true,
     onNavigationIconClick: () -> Unit = {},
+    onLinkClick: (url: String, packageName: String?) -> Unit = { _, _ -> },
```
takahirom, line 52: I think we can make a sealed interface or class for this. It makes the app more flexible.
But I will make a issue for it 👍

takahirom:
LGTM!

feature-about/src/main/java/io/github/droidkaigi/confsched2022/feature/about/About.kt
fornewid: You're right!
Actually I want to know opinions about this.
Now someone will follow this to another issue. 🤗

PR #339, opened 2022-09-10
Migrate SessionDetail to use UiLoadState instead

kittinunf:
## Issue
- close #ISSUE_NUMBER
- N/A work on foundation

## Overview (Required)

After the successful proposal of #287, we start moving the current code to use UiLoadState instead. This should not have any changes in the UI

## Screenshot

Before | After
:--: | :--:
 |


takahirom (approved):
Thanks!

PR #350, opened 2022-09-10
Migrate to use UiLoadState for the rest of the screens

kittinunf:
## Issue
- N/A - a foundation work

## Overview (Required)

A follow-up PR for #287 and #339, move the Session screen to use `UiLoadState` instead.

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
Thank you for simplify the DroidKaigi app!

PR #351, opened 2022-09-10
Calculate the height from the original time minus 1 minutes and add end padding

### LINKED ISSUE #345: Reduce 1 minute when calculate timetable item height to separate each sessions.
**Idea Description**
We want to make a space here so. When calculating session height, I would like to use `minute - 1` height.
image

https://github.com/DroidKaigi/conference-app-2022/blob/60ab6c88adf362103600a05d8b2a386d60a51f53/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt#L252

rmakiyama:
## Issue
- close #345

## Overview (Required)
- As suggested, Calculate the height from the original time minus 1 minutes.
- ~💡In my opinion, I thought `TimetableItem` would be nice to have end padding, so I added it.~

~If you don't want the end padding, let me know so I can remove it.~ remove it 👍🏻
See https://github.com/DroidKaigi/conference-app-2022/pull/351#pullrequestreview-1103130117

## Screenshot
Before | After | After(include my idea) |
:--: | :--: | :--:
 |  |

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
@@ -56,7 +56,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.launch
 import kotlinx.datetime.DateTimeUnit
-import kotlinx.datetime.DateTimeUnit.Companion
 import kotlinx.datetime.Instant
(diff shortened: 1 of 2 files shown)


takahirom (approved):
Nice to remind the room padding! But we have the design that is a show the line and align the session to the line.
Your changes remind me that all we have to do is put padding 1dp to the start.
I'll make the issue. So can you remove the padding to the end?

rmakiyama:
> Your changes notice me that all we have to do is put padding 1dp to the start.

Sounds good 👏🏻 I will remove end padding, and push it.

Issue #347, opened 2022-09-10
Move session detail date to strings.xml's format


takahirom:
**Idea Description**
We have English speakers at the conference. So we have to show English. So please adapt by strings.xml resource.

**Reference images and links**
strings.xml format
https://developer.android.com/guide/topics/resources/string-resource#formatting-strings

https://github.com/DroidKaigi/conference-app-2022/blob/b596f1baa82f345d1645a94b6c9f8f5f685b176f/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt#L312

att55:
I'd like to do this task 🙋🏼

takahirom:
Thanks! Assigned!

PR #359, opened 2022-09-10
Use formatting strings in session date

att55:
## Issue

- close https://github.com/DroidKaigi/conference-app-2022/issues/347

## Overview (Required)

Use formatting strings in session date for English ver.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature-sessions/src/main/res/values/strings.xml
+++ b/feature-sessions/src/main/res/values/strings.xml
@@ -2,7 +2,6 @@
 <resources>
     <!-- Session Detail Screen -->
     <string name="session_description_read_more_text">続きを見る</string>
-    <string name="session_schedule">%1$s %2$s %3$s-%4$s</string>
     <string name="session_target_audience">対象者</string>
     <string name="session_speaker">スピーカー</string>
     <string name="session_material">資料</string>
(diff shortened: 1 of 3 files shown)


feature-sessions/src/main/res/values/strings.xml
```diff
@@ -2,6 +2,7 @@
 <resources>
     <!-- Session Detail Screen -->
     <string name="session_description_read_more_text">続きを見る</string>
+    <string name="session_schedule">%1$s %2$s %3$s-%4$s</string>
```
takahirom, line 5: I think you can add 月 日 here. I think if you do this, you don't need if expression in the compose. 🙏

att55: Are you ok an expression in English as bellow?

```
10 5th 10:00-11:00
```

I used if expression for capitalized month name in English.

```
October 5th 10:00-11:00
```

takahirom: I see probably I made a mistake about this. Probably we mustn't use strings.xml. Because we will have to manage two place.
Can you remove it from strings.xml and make date string in Kotlin?

att55: > Can you remove it from strings.xml and make date string in Kotlin?

OK. I will try it!

att55: @takahirom 
Fixed!
Please check when you have the time. 🙏🏼

takahirom (approved):
Thank you for your cooperation! LGTM!

PR #366, opened 2022-09-11
[Proposal] Move build logic plugins into gradle/plugins folder

myoshita:
## Issue
- None

## Overview (Required)
Since gradle 7.5.1, recommended that putting all gradle details as subfolder of gradle. 
This change is applied to [idiomatic-gradle](https://github.com/jjohannes/idiomatic-gradle) on which now in android app is based.
The purpose of this approach is described to following:

> - Putting it as subfolder of gradle:
I think it is good practice to put as much gradle-related files into the gradle folder instead of spamming the root with gradle related files and folders. This is also an advantage over 'buildSrc' which is hard-coded in the root. Other build configuration, like the Version Catalog, is also located in gradle by default
> - Calling it plugins:
In many samples, the folder is called build-logic (or some variation of that). By now, I think plugins is a better name because that is the term for what folks will put into this folder - local Gradle plugins. build-logic on the other hand is yet another term they have to grasp. Also the term 'logic' is debatable if you only define convention plugins (which are more build-configuration than build-logic imo).

I think too that it is good idea to gather gradle files into gradle folder. (like version catalog) 
But, I'm concerned that it may be difficult to find where the plugin is located than build-logic.

## Links
- https://github.com/gradle/gradle/pull/21192
- https://github.com/jjohannes/idiomatic-gradle/commit/b7299907cbf81af3b08768052ef79fe4cd9071a3
- https://github.com/android/nowinandroid/tree/main/build-logic

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
👀

takahirom:
I didn't know that. 👀  May I ask how you knew this?

takahirom (approved):
Thank you for updating the DroidKaigi app base structure!

myoshita:
Thank you for reviewing this PR!

>I didn't know that. 👀 May I ask how you knew this?

I found out that recently when looked repositories (such as nowinandroid, idiomatic-gradle) to write documentation for my project.

PR #367, opened 2022-09-11
[iOS] explicit running thread when collecting coroutine flow

RyuNen344:
## Issue
- close nothing

## Overview (Required)
- iOS app crash when launching

```log
2022-09-11 15:28:47.204205+0900 DroidKaigi2022[17747:5918236] *** Terminating app due to uncaught exception 'NSGenericException', reason: 'Calling Kotlin suspend functions from Swift/Objective-C is currently supported only on main thread'
```

- I put `DispatcheQueue.main` to explicit running thread
- I don't know what is the best, so please tell me if more beautiful deal

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Thanks!

I thought `@MainActor` would work, but after trying it, it doesn't seem to work. So, I hope to improve this at another time.

PR #370, opened 2022-09-11
Enable Kotlin Explicit API Mode (Warning)

### LINKED ISSUE #279: Introduce Kotlin Explicit API mode to core-xxx modules
**Idea Description**

Explicit visibility using `public` or `internal` or `private` allows you to control what expose outside the module. Kotlin default visibility is `public`, so you may expose things you don't want to be public. 

You can force explicit visibility using [Kotlin Explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md). It helps control visibility.

⚠️ Applying Explicit API mode requires changing many parts of existing code.

d1-morimoto:
## Issue
- close #279

## Overview (Required)
- Kotlin Explicit API Mode warnings will display when editing core-xxx modules.

## Links
- [Kotlin Explicit API mode](https://github.com/Kotlin/KEEP/blob/master/proposals/explicit-api-mode.md)

## Screenshot
Before | After
:--: | :--:
before | after

### DIFF
--- a/core-data/build.gradle.kts
+++ b/core-data/build.gradle.kts
@@ -11,6 +11,8 @@ plugins {
 android.namespace = "io.github.droidkaigi.confsched2022.core.data"
 
 kotlin {
+    explicitApiWarning()
+
     sourceSets {
         val commonMain by getting {
             dependencies {
--- a/core-designsystem/build.gradle.kts
+++ b/core-designsystem/build.gradle.kts
@@ -6,6 +6,8 @@ plugins {
 android.namespace = "io.github.droidkaigi.confsched2022.core.designsystem"
 
 kotlin {
+    explicitApiWarning()
+
     sourceSets {
         val commonMain by getting {
         }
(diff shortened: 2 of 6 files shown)


d1-morimoto:
I made this for now.
I am concerned that the following modules do not display warnings, but I can't think of a reason why.

* core-testing
* core-ui

I would appreciate it if you could give me some suggestions for modifications through your review:bow:

takahirom:
Probably because they are android module, right?

d1-morimoto:
Oh, that's it!
Thanks for the advice, I realized my mistake:bow:

d1-morimoto:
I believe it has been corrected.
I would be happy to hear if there is anything else that needs to be done:bow:

* core-data
    core-data
* core-designsystem
    core-designsystem
* core-model
   core-module 
* core-testing
    core-testing
* core-ui
    core-ui
* core-zipline
    core-zipline

takahirom (approved):
Thanks! LGTM 👍

d1-morimoto:
@takahirom 
Thanks for the review!

PR #386, opened 2022-09-12
[iOS] Disallow header comments

woxtu:
## Issue
- N/A

## Overview (Required)
- This PR enables the `file_header` rule to disallow header comments except workaround comments.

## Links
- https://realm.github.io/SwiftLint/file_header.html

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Good catch! Thank you!

Issue #259, opened 2022-09-07
Introduce multiplatform resource


takahirom:
**Idea Description**

I would like to manage strings.

I tried moko-resource, but I couldn't solve the CI build error.

https://github.com/DroidKaigi/conference-app-2022/pull/119

limitation

* We can use it in Kotlin Multiplatform
* We can change it by locale. (EN/ JP etc)

**Reference images and links**

kittinunf:
@takahirom The reason why it didn't work on CI because we use Ubuntu for our CI, moko-resouces throws error when the host is not macOS.

I did that in the past in my prev. company, when we change the host to macOS, it works well. 

Can we change the host to mac instead? If not, we won't be able to resolve the CI checks.

takahirom:
I didn't know that 👀 
We can use GitHub Actions as much as we'd like. Because It's free for open source. 😄

takahirom:
@kittinunf Do you want to try that?

kittinunf:
I definitely can since I have done it before! But if we don't have macOS CI for github action (I think it is only in the paid plan), I don't think it is worth without CI checks.

kittinunf:
In my experience, we shared Strings and Colors, it wasn't working with the Image resources though (but that was version 0.18.0 which was quite some time ago).

takahirom:
I'm not sure if this problem is related to the host OS because the error message is `multiplatform_strings.xml: Error: Duplicate resources`.
I thought it is not related to the host 👀 
But probably we will need to try to build on Mac OS for iOS. That's for sure.
https://github.com/DroidKaigi/conference-app-2022/pull/119

kittinunf:
How about if you could setup a macOS host for CI GitHub, I could help you with this one, what do you think?

takahirom:
Do you mean we change this to `runs-on: macos-12` ?
https://github.com/DroidKaigi/conference-app-2022/blob/main/.github/workflows/Build.yml

kittinunf:
👍 Thanks sounds like we can move ahead with this.

kittinunf:
@takahirom @ry-itto I couldn't build the iOS project. It says that AssetColors file is not found (not generated by SwiftGen plugin).

Do you have any tick to make iOS working and/or runnable?

kittinunf:
Ah, I got it, we need to run scripts under `app-ios/` folder first.

ry-itto:
@kittinunf 
[...]

PR #387, opened 2022-09-12
Introduce MokoResources for the project

kittinunf:
## Issue
- #259 but it can't close the issue just yet because we have to make sure it works on iOS.

## Overview (Required)
- In this PR, we are introducing the moko-resource to maintain/manage our string files.

Now I have only migrated the about section first. First, if this can move forward, I would move on to the rest of the app.

## Links
-

## Screenshot
In Jp:

Screen Shot 2022-09-12 at 12 39 48

In En:

Screen Shot 2022-09-12 at 12 44 55


core-model/build.gradle.kts
```diff
@@ -22,4 +23,15 @@ kotlin {
             }
         }
     }
-}
\ No newline at end of file
+}
+
+dependencies {
```
kittinunf, line 28: @takahirom I know this is questionable but I need a place where we could satisfy the direct exposure to the ios-combined module. Now the options are;

```
val projects = listOf(
            projects.coreModel,
            projects.coreDesignsystem,
            projects.coreData
        )
```

I found `coreModel` to be the sanest one.

core-model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/strings/ResExt.kt
```diff
@@ -0,0 +1,7 @@
+package io.github.droidkaigi.confsched2022.strings
+
+import io.github.droidkaigi.confsched2022.model.Res
+import io.github.droidkaigi.confsched2022.model.Res.strings
+
+val Strings: strings
```
kittinunf, line 6: Find this to be easier than spell it out the whole thing :)

core-model/build.gradle.kts
takahirom: I don't dislike this idea. 👀 
But originally, I added core resources.
I have a question about this. Can we use string value in the model class?

https://github.com/DroidKaigi/conference-app-2022/pull/119/files#diff-c053860dd51d0a036f0a1ffc0adbca71dd1dcf8f9610e502da3e1c1f02db1f68R12

By the way, can you move it here?

```
kotlin {
    sourceSets {
        val commonMain by getting {
            dependencies {
// here

kittinunf: Sure I can! 

I don't see a reason why we can't use the string value in the model class. It should be possible.

takahirom:
I checked it in my local 👍  It's great! Let's merge then. Can you resolve the conflict?

kittinunf:
Yes! I will sorry got swamped with work stuff. I will work on this and resolve it tonight 💦

takahirom:
> I will help refactor to use moko resource in all others screens later.

Thank you!

PR #390, opened 2022-09-12
[iOS] Use @MainActor instead of DispatchQueue.main

ShotaKashihara:
## Issue
- N/A

## Overview (Required)
- This PR is another idea to improve [this patch](https://github.com/DroidKaigi/conference-app-2022/pull/367)
- I suggest using `@MainActor` instead of `DispatchQueue.main`.
- No crashes have occurred, but there may yet be a better way 😄

## Links
-

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Looks great! Let's wait for ry-itto

ry-itto (approved):
Thank you for improving!
This is what I wanted to do 🙇

PR #398, opened 2022-09-12
Fix contributors screen accessibility

yukilabo:
## Overview (Required)
- Fix TalkBack says contributors name twice.

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
Thanks!

PR #401, opened 2022-09-13
Remove searchContentHeight

### LINKED ISSUE #325: Remove searchContentHeight
**Idea Description**
A ordinary variable should not be used in Composable function if you want to recompose.
And I think we can do it with wrap content.

https://github.com/DroidKaigi/conference-app-2022/blob/main/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt#L190

hiesiea:
## Issue
- close #325 

## Overview (Required)
- Remove searchContentHeight in SearchedItem

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
Thank you for your refactoring!

Issue #408, opened 2022-09-13
[edge-to-edge] Draw content behind the navigation bar


Mori-Atsushi:
**Idea Description**
It is recommended to draw content behind the navigation bar in Android 10 or later.
This allows content to be drawn from edge-to-edge of the screen, enhancing the immersive experience of the application.

**Reference images and links**
https://developer.android.com/develop/ui/views/layout/edge-to-edge

Mori-Atsushi:
I can tackle the first screen 🙋‍♂

takahirom:
Thanks as always! Assigned 🙏

PR #414, opened 2022-09-13
[edge-to-edge] Draw content behind the navigation bar (partial)

Mori-Atsushi:
## Issue
- #408

## Overview (Required)
- Allow content to be drawn behind the navigation bar.
- Change `KaigiScaffold` to pass inner padding to child Composable.
- Since there is no automatic coloring of navigation bar under android 10, specify a translucent color for the navigation bar.
- Support other than the About screen is partial.

## TODO
- Fully support other screens.

## Screenshot
Before | After
:--:|:--:
Screenshot_20220913_192848|Screenshot_20220913_192555
Screenshot_20220913_192912|Screenshot_20220913_192451

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2022/MainActivity.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2022/MainActivity.kt
@@ -21,6 +21,7 @@ class MainActivity : ComponentActivity() {
             if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
                 Color.TRANSPARENT
             } else {
+                // Make it easy to see navigation buttons for older OS versions
                 Color.argb((255 * 0.5).toInt(), 0, 0, 0)
             }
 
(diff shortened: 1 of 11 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2022/MainActivity.kt
```diff
@@ -14,6 +16,13 @@ class MainActivity : ComponentActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         WindowCompat.setDecorFitsSystemWindows(window, false)
+        window.statusBarColor = Color.TRANSPARENT
+        window.navigationBarColor =
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.Q) {
+                Color.TRANSPARENT
+            } else {
+                Color.argb((255 * 0.5).toInt(), 0, 0, 0)
```
takahirom, line 24: // Make it easy to see for older OS versions

Mori-Atsushi: Thank you. I did it!

takahirom (approved):
Looks great!

PR #416, opened 2022-09-13
Translated into Simplified Chinese.

Hamachi-geek:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- The app was translated into Simplified Chinese.


takahirom:
Thanks as always!
I would like to merge this first. Because it will cause a lot of conflict🙇
https://github.com/DroidKaigi/conference-app-2022/pull/387

takahirom:
Can you take a look at the conflict?

feature-about/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kittinunf, line 1: This should be moved to `core.model/commonMain/resources`

Hamachi-geek: In "MR" directly right?

kittinunf: Yes!

feature-announcement/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kittinunf, line 1: This should move into `core-model` resources.

feature-contributors/src/main/res/values-zh-rCN/strings.xml
```diff
@@ -0,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kittinunf, line 1: Same here.

Hamachi-geek: Sorry. I understood what I have to do right now.

kittinunf:
Or you can rebase on top of #482. It has everything you need except for the translation text itself.

takahirom (approved):
Thank you for adding the translation!

Issue #405, opened 2022-09-13
Change KaigiTopAppBar color 


takahirom:
**Idea Description**
app
<img width="470" alt="image" src="https://user-images.githubusercontent.com/1386930/189853765-84a962d4-96d6-46c4-9cbd-b0fc4a12acd5.png">
design
<img width="242" alt="image" src="https://user-images.githubusercontent.com/1386930/189853968-9ec89af1-3494-4e67-81b2-09a8ffb49da2.png">

**Reference images and links**

itochan:
I want to try.
Please assign to me.

att55:
I'd like to do this 🙋🏼

takahirom:
Thanks! Assigned 👍

PR #418, opened 2022-09-13
Unify all app bar colors

att55:
## Issue
- close https://github.com/DroidKaigi/conference-app-2022/issues/405

## Overview (Required)

Unify all app bar colors.


takahirom (approved):
I know It's much more difficult than it seems! Thanks!

PR #446, opened 2022-09-14
:recycle: Update Language tags are now displayed.

### LINKED ISSUE #335: [API]I want to be able to display the supported languages for the session as well as the web version.
**Idea Description**
On the web, it seems to be possible to display JP,EN or EN,JP, etc.
However, the API response structure used in the application only includes the speaker's first language, so it is not possible to support the same display as the web version.
To solve this, I think the structure should be languages(Array[String]) instead of language(String).

スクリーンショット 2022-09-10 7 21 50
スクリーンショット 2022-09-10 7 22 11

Corvus400:
## Issue
- close #335 

## Overview (Required)
- Language tags are now displayed in two places: session details and session list.
  - The colors of the tags are the same as in the web version, but I believe they should be changed in the app as well, once the design is defined.

## Links
- https://droidkaigi.jp/2022/timetable

## Screenshot
Before | After
:--: | :--:
 | 
 | 
 | 
 | 
 |


takahirom:
It's definitely useful!

takahirom:
Sorry, let me contact with Designer.

takahirom:
Looks too colorful. Can you use the 20min color for languages?
image

Corvus400:
@takahirom 
Changed as shown in the image.😃👍

takahirom:
Looks good! Thank you for your cooperation!

PR #454, opened 2022-09-14
Delete needAuth check

### LINKED ISSUE #442: Authenticated even if NetworkService auth is false.
**Idea Description**
Even if I set the needAuth property of the get function of the NetworkService class to false, Authorization: Bearer ***** is still added as a header.

example:
https://github.com/DroidKaigi/conference-app-2022/blob/main/core-data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/staff/StaffApi.kt#L16
https://github.com/DroidKaigi/conference-app-2022/blob/main/core-data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/NetworkService.kt#L18-L29

momomomo111:
## Issue
- close #442 

## Overview (Required)
- This app don't have the API that doesn't need auth. Fixed so that it is always true.


takahirom (approved):
Thank you for cleaning the code!

PR #461, opened 2022-09-14
Apply edge to edge into contributors page

### LINKED ISSUE #458: [edge-to-edge] Apply edge to edge into contributors page
**Idea Description**
On the #408, just applied the edge to edge into the about page. But we should apply an edge to edge all other pages too.
In this issue, we focus on the contributors page.

https://developer.android.com/develop/ui/views/layout/edge-to-edge
https://github.com/DroidKaigi/conference-app-2022/issues/408

chigichan24:
## Issue
- close #458

## Overview (Required)
- I applied edge to edge into the contributors page.
- Pass the `innerPadding` as `contentPadding` in `LazyColumn`.

## Links
- https://developer.android.com/develop/ui/views/layout/edge-to-edge

## Screenshot
Description | Before | After
:--: | :--: | :--:
3-buttons case(portrait) |  | 
gesture navigation case(portrait) |  | 
3-buttons case(landscape) |  | 
gesture navigation case(landscape) |  |


takahirom:
👀

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
```diff
@@ -59,7 +58,7 @@ fun Contributors(
             )
         }
     ) { innerPadding ->
-        Box(modifier = Modifier.padding(innerPadding)) {
+        Box(modifier = modifier) {
             when (uiModel.state) {
                 is Error -> TODO()
                 Loading -> Box(
```
Mori-Atsushi, line 64: I think the loading mark should be centered along with the other screens, so I prefer to add window insets padding here as well!

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
```diff
@@ -59,7 +58,7 @@ fun Contributors(
             )
         }
     ) { innerPadding ->
-        Box(modifier = Modifier.padding(innerPadding)) {
+        Box(modifier = modifier) {
```
takahirom, line 61: The modifier parameter should not be passed to child composable. But it was originally so. We can make a issue and you can leave it as it is.

takahirom (commented):
Please check mori san's review🙏

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
chigichan24: @Mori-Atsushi @takahirom change: [912214e](https://github.com/DroidKaigi/conference-app-2022/pull/461/commits/912214efa44a1e6e78facb8d2ce93e431b9ba8b1)  I have no idea except passing innerPadding to the loading indicator. What do you think? If you have any further nice solutions please tell me.

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
chigichan24: I believe that it is better to pass on the necessary variables only when they are needed. In this case, `Modifier` is the same situation. So I clean up the modifier as soon as I stopped passing it to grandchildren composable.

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
Mori-Atsushi: That looks good to me!

takahirom: I think we should always pass modifiers.
We are having the lint for it. So, for now, it is ok. If you want to fix this. You can make a PR!
https://chris.banes.dev/always-provide-a-modifier/

feature-contributors/src/main/java/io/github/droidkaigi/confsched2022/feature/contributors/Contributors.kt
takahirom: I think we should always pass modifiers.
We are having the lint for it in https://github.com/DroidKaigi/conference-app-2022/pull/464. So, for now, it is ok. If you want to fix this. You can make a PR!
https://chris.banes.dev/always-provide-a-modifier/

PR #463, opened 2022-09-14
[edge-to-edge] Apply edge to edge into SessionsScreen

### LINKED ISSUE #459: [edge-to-edge] Apply edge to edge into SessionsScreen
**Idea Description**
* I think SessionsScreen also prefers to render content behind the navigation bar.
* We also need to fix the issue that the navigation bar and buttons overlap in landscape.

portrait|landscape
:--:|:--:
Screenshot_20220914_214302|Screenshot_20220914_214335

https://developer.android.com/develop/ui/views/layout/edge-to-edge
https://github.com/DroidKaigi/conference-app-2022/issues/408

Mori-Atsushi:
## Issue
- close #459

## Overview (Required)
- I applied edge to edge into SessionsScreen.
- I fixed the issue that the navigation bar and buttons overlap in landscape
- Extends the bounds of the TimetableScreen to allow scrolling to the bottom

## Screenshot
Before | After
:--: | :--:
Screenshot_20220914_234455|Screenshot_20220914_233841
Screenshot_20220914_234520|Screenshot_20220914_234007
Screenshot_20220914_234503|Screenshot_20220914_233946
Screenshot_20220914_234548|Screenshot_20220914_233918

### DIFF
--- a/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
+++ b/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
@@ -7,6 +7,7 @@ import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.transformable
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.WindowInsets
 import androidx.compose.foundation.layout.WindowInsetsSides
@@ -141,7 +142,6 @@ fun Sessions(
         Column(
             modifier = Modifier
                 .padding(top = 4.dp)
-                .padding(innerPadding)
         ) {
             when (scheduleState) {
                 is Error -> {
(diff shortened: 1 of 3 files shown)


takahirom:
Looks good. Let me take a look tomorrow 🙏

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -136,15 +145,16 @@ fun Sessions(
...
-                .padding(innerPadding)
         ) {
             when (scheduleState) {
                 is Error -> {
                     scheduleState.value?.printStackTrace()
                     TODO()
                 }
                 Loading -> Box(
-                    modifier = modifier.fillMaxSize(),
+                    modifier = modifier
```
takahirom, line 155: We should fix this after. 👀  (I would like to use Modifier)

feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
```diff
@@ -548,8 +568,9 @@ private class TimetableScreen(
             } else {
                 0f
             },
-            maxY = if (height < timetableLayout.timetableHeight) {
-                -(timetableLayout.timetableHeight - height).toFloat()
+            maxY = if (height < timetableLayout.timetableHeight - bottomPaddingPx) {
```
takahirom, line 571: 👍

takahirom:
This is definitely better! So I was just wondering which to use, consuming the window insets or passing the contentPadding parameter. Is there any opinion about his?

Mori-Atsushi:
I'm wondering about that too.

Since Scaffold is strongly recommended to use inner padding (lint throws an error if you don't use), I used this.
But I think it's clearer to use Window Insets directly.
I don't really like Scaffold that does too much 😂

Issue #407, opened 2022-09-13
Try Twitter Compose rule


takahirom:
**Idea Description**
Our repository can be a playground for this 👍 
This issue is just a trial, so it is possible that it may not make it to the merge into the main branch.

**Reference images and links**
https://github.com/twitter/compose-rules

takahirom:
Can we use detectRule?

https://github.com/DroidKaigi/conference-app-2022/blob/main/gradle/plugins/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/DetektPlugin.kt
We can check it by `./gradlew composeLint`

kwmt:
I'm interested in the issue. Let me try!

sobaya-0141:
Thanks! Assigned :+1:

PR #464, opened 2022-09-14
added twitter compose rules with detekt

kwmt:
## Issue
- close https://github.com/DroidKaigi/conference-app-2022/issues/407

## Overview (Required)
- In this PR I disable [TwitterCompose settings in the detekt.yml](https://github.com/DroidKaigi/conference-app-2022/pull/464/files#diff-5773362839b175cf54bf2e485eba26e91827308200695eb1b592a1f491f0bd2aR27-R53), but I confirmed the rule works fine.
- To check this rules, you need to enable TwitterCompose settings in the detekt.yml and then you need to run `./gradlew composeLint`.

Fcn45lcagAAEysv

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/gradle/plugins/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/ProjectGradleDsl.kt
+++ b/gradle/plugins/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/ProjectGradleDsl.kt
@@ -38,4 +38,10 @@ private fun DependencyHandlerScope.api(
     add("api", artifact.get())
 }
 
+fun DependencyHandlerScope.detektPlugins(
+    artifact: Optional<Provider<MinimalExternalModuleDependency>>
+) {
+    add("detektPlugins", artifact.get())
+}
+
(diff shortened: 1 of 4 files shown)


gradle/plugins/src/main/kotlin/io/github/droidkaigi/confsched2022/primitive/ProjectGradleDsl.kt
```diff
@@ -38,4 +38,10 @@ private fun DependencyHandlerScope.api(
     add("api", artifact.get())
 }
 
+fun DependencyHandlerScope.detektPlugins(
```
kwmt, line 41: Perhaps it don't need this method.

takahirom:
@kwmt I think we can merge this 👀 What do you think?

kwmt:
@takahirom Sorry for the delay.
OK, I wrote a description and I did ready the PR. Could you please check this and merge?

takahirom:
We can start now!

takahirom:
Thanks!

kwmt:
Thank you for your review and merge 😄

PR #479, opened 2022-09-15
[iOS] Open License Page from AboutView

### LINKED ISSUE #429: [iOS] Open License Page from `AboutView`
**Idea Description**

- Implement license page ([LicensePlist](https://github.com/mono0926/LicensePlist) way is okey)
- Open license page from `AboutView`

touyou:
## Issue
- close #429 

## Overview (Required)
- Open License page from AboutView
- I used [LicenseList](https://github.com/cybozu/LicenseList) library because [LicensePlist](https://github.com/mono0926/LicensePlist) cannot generate SPM dependencies. ([ref](https://blog.cybozu.io/entry/2022/07/05/173000))
  - I wrote package plugin of LicenseList, but it's not officially supported by LicenseList. I'm offering support in https://github.com/cybozu/LicenseList/issues/4

### For reviewer

- I implemented navigations only by standard library, but I'm not sure it is good way.
  - I also found this library. https://github.com/johnpatrickmorgan/TCACoordinators
  - Honestly, I want to use `NavigationStack` 😂

## Links
- Nothing

## Screenshot

https://user-images.githubusercontent.com/465697/190327609-da0ebee6-08b6-454e-9088-9d91c018aa81.mp4


ry-itto (commented):
How about adding `app-ios/Sources/AboutFeature/license-list.plist` file to .gitignore ?

I think it is generated file by SPM Plugin. So I think it is not needed to manage with git.

touyou:
> How about adding app-ios/Sources/AboutFeature/license-list.plist file to .gitignore ?

I understand and agree, but LicenseList program cannot generate file properly without template file.
I'll try to generate template file in plugin code 👍

touyou:
I tried it. But It seems that Xcode cannot recognize the file generated during plugin execution phase that didn't exist originally.
So if we run the code, app crash only at first time.

Anyone does have some solutions of this?

ry-itto:
Hmm...
It's okey! Let's try with another issue or PR :pray:

ry-itto (approved):
LGTM 👍🏼 
Thanks!

PR #462, opened 2022-09-14
[Foundation] Migrate to StringResource for Sponsor and Staff screen

kittinunf:
## Issue
- N/A, this PR is a continuation from #387, #441, and #450

## Overview (Required)
I have migrated 2 more screens into this setup Sponsor and Staff screen.

## Links
-

PS. I will have another last PR to migrate app/android into this setup as well.

## Screenshot
Before | After
:--: | :--:
 |

PR #450, opened 2022-09-14
[Foundation] Migrate to StringResource for map, sessions and setting screen

kittinunf:
## Issue
- N/A, this PR is a continuation from #387, #441

## Overview (Required)
I have migrated 2 more screens into this setup Sessions and Settings screen.

## Links
-

PR #441, opened 2022-09-14
[Foundation] Add string for announcement and contributor

kittinunf:
## Issue
- N/A continuation from #387 

## Overview (Required)
I have migrated 2 more screens into this Announcement and Contributor.

## Screenshot

Before | After
:--: | :--:
 | 

In 日本語

In English

PR #482, opened 2022-09-15
[Foundation] Migrate app-android to Moko-resources

kittinunf:
## Issue
- N/A, this PR is a continuation from #387, #441, #450 and #462

## Overview (Required)
I have migrated the last module to moko-resources.

## Links
-

This would conclude the work for Android. I will start working on iOS next.


core-model/src/commonMain/resources/MR/zh-rCN/strings.xml
```diff
@@ -0,0 +1,49 @@
...
+    <string name="title_sessions">动态消息</string>
+    <string name="title_about">关于DroidKaigi</string>
+    <string name="title_announcement">通知</string>
+    <string name="title_map">楼面导轨</string>
+    <string name="title_sponsors">赞助者</string>
+    <string name="title_contributors">Contributor</string>
+    <string name="title_setting">设置</string>
+
+    <!-- About -->
+    <string name="about_top_app_bar_title">DroidKaigiとは</string>
```
takahirom, line 13: I have a question! If we don't have these strings, it will fail to build?

kittinunf: Our lint would prevent our build on the CI to pass.

takahirom: I see. If I use it in production. We should consider it. 👀 
https://github.com/icerockdev/moko-resources/issues/331

PR #502, opened 2022-09-16
Suppress warnings in build.gradle.kts (DroidKaigi2022)

### LINKED ISSUE #438: 'val Project.libs: LibrariesForLibs' can't be called in this context by implicit receiver. Use the explicit one if necessary
**Describe the bug**

**To Reproduce**

**Expected behavior**

**Screenshots**

tatsuyafujisaki:
## Issue
- close #438

## Overview (Required)
- Suppress the warning `val Project.libs: LibrariesForLibs' can't be called in this context by implicit receiver. Use the explicit one if necessary` in build.gradle.kts (DroidKaigi2022).
  - First, I tried to add `@Suppress("DSL_SCOPE_VIOLATION")` above `plugins { … }` as follows …
    
… but got the build error `e: build.gradle.kts:1:33: Expecting an expression`.

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 plugins {
+    // https://youtrack.jetbrains.com/issue/KTIJ-19369/False-positive-cant-be-called-in-this-context-by-implicit-receiver-with-plugins-in-Gradle-version-catalogs-as-a-TOML-file#focus=Comments-27-6204464.0-0
     @Suppress("DSL_SCOPE_VIOLATION")
     alias(libs.plugins.androidGradlePlugin) apply false
     @Suppress("DSL_SCOPE_VIOLATION")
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Thanks!

build.gradle.kts
```diff
@@ -1,6 +1,9 @@
 plugins {
+    @Suppress("DSL_SCOPE_VIOLATION")
```
takahirom, line 2: Can you add this comment to tell readers why we need this? 👀 
// https://youtrack.jetbrains.com/issue/KTIJ-19369/False-positive-cant-be-called-in-this-context-by-implicit-receiver-with-plugins-in-Gradle-version-catalogs-as-a-TOML-file#focus=Comments-27-6204464.0-0

tatsuyafujisaki: https://github.com/DroidKaigi/conference-app-2022/pull/502/commits/231a5830dad6932487a777594b8f438119b84873 Sure. I did.:+1:

PR #503, opened 2022-09-16
update zh-CN strings.xml

Nthily:
# Overview (Required)
- as the title says


takahirom (approved):
Thanks for your support!

PR #456, opened 2022-09-14
Modify the real session on the JS side to make Zipline validatable

kubode:
## Issue
- Part of #250

## Overview (Required)
- To validate the zipline on the Android app, modify the real session.

## Screenshot

No screenshots.

PR #505, opened 2022-09-16
Fix Zipline is not working as well

### LINKED ISSUE #250: Zipline is hanging when calling
**Idea Description**

When we call it, the response does not come. When we use a local server, It works.

https://github.com/DroidKaigi/conference-app-2022/blob/9a5093652a62da0c96e7914be6702cee5c56f1b0/core-zipline/src/androidMain/kotlin/io/github/droidkaigi/confsched2022/zipline/SessionsZipline.kt#L90

kubode:
## Issue
- close #250 

## Overview (Required)
- See https://github.com/DroidKaigi/conference-app-2022/issues/250#issuecomment-1248231463

## Another issue

I have made changes to the lunch session at https://github.com/DroidKaigi/conference-app-2022/pull/456, but they are not reflected in the `SessionDetailScreen` of the lunch session.
This is because the `SessionDetailViewModel` does not use `SessionsZipline`.
Also, the `ScheduleModifier` can change the `DroidKaigiSchedule`, but it cannot change the individual `TimetableItem` that the `SessionDetailViewModel` references.

To apply Zipline to `SessionDetailScreen` as well, we need to re-architect such as integrating Zipline into the `:core-data` module.


kubode:
I'm working on fixing a failing test.
I checked the log of the test and found the following error.


```
Caused by: java.lang.UnsatisfiedLinkError: no quickjs in java.library.path: [/Users/masatoshi/Library/Java/Extensions, /Library/Java/Extensions, /Network/Library/Java/Extensions, /System/Library/Java/Extensions, /usr/lib/java, .]
	at java.base/java.lang.ClassLoader.loadLibrary(ClassLoader.java:2673)
	at java.base/java.lang.Runtime.loadLibrary0(Runtime.java:830)
	at java.base/java.lang.System.loadLibrary(System.java:1873)
	at app.cash.zipline.QuickJsNativeLoaderKt.loadNativeLibrary(QuickJsNativeLoader.kt:19)
	at app.cash.zipline.QuickJs.<clinit>(QuickJs.kt:35)
	at app.cash.zipline.Zipline$Companion.create(Zipline.kt:169)
	at app.cash.zipline.loader.ZiplineLoader.loadFromManifest$zipline_loader_debug(ZiplineLoader.kt:235)
	at app.cash.zipline.loader.ZiplineLoader$load$1$1$1.invokeSuspend(ZiplineLoader.kt:175)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

I looked for an issue in `cashapp/zipline`, but there did not seem to be any related issues.
I get the same error on Linux.

Do you know of a workaround for this error?

takahirom:
👀

takahirom:
Probably we used to use try-catch to load the library for unit test 👀

kubode:
I found a workaround.

Adding this solves the error.
```
dependencies {
    testImplementation("app.cash.zipline:zipline-jvm:0.9.4")
}
```


Perhaps the `zipline-android` engine is used by default on unit tests.
I think that the test on Robolectric runs in the JVM, so it needs the `zipline-jvm` engine.

takahirom:
It is more ideally. But I don't want to use a real server for unit tests due to test flakiness.

takahirom:
But I think complete is better than perfect 😄 
You can use JVM implementation for now!

kubode:
> It is more ideal. But I don't want to use a real server for unit tests due to test flakiness.

Agree.

What about changing `SessionsZipline` to an interface, injecting an implementation class to ViewModels, and injecting the fake class during testing?

takahirom:
It's what I was thinking about! It is the best way I have come up with.

kubode:
Fixed.
Is the location and naming of components appropriate?

takahirom:
👀

core-zipline/src/androidMain/kotlin/io/github/droidkaigi/confsched2022/zipline/SessionsZipline.kt
```diff
@@ -62,43 +63,50 @@ class SessionsZipline @Inject constructor(
...
         }
-        val modifierStateFlow = MutableStateFlow(defaultModifier)
+    }
 
-        coroutineScope.launch(dispatcher) {
-            var zipline: Zipline? = null
+    @OptIn(ExperimentalTime::class) // measureTimedValue
+    override fun timetableModifier(): Flow<suspend (DroidKaigiSchedule) -> DroidKaigiSchedule> {
+        return channelFlow {
+            // The JS modifier takes about 300 ms to execute.
```
takahirom, line 89: The JS modifier -> Loading of JS modifier
This can lead misunderstanding👀

takahirom (commented):
Great! Can you take a look at small points?

core-zipline/src/androidMain/kotlin/io/github/droidkaigi/confsched2022/zipline/SessionsZipline.kt
kubode: Improved this comment.
Please check [05cfe3b](https://github.com/DroidKaigi/conference-app-2022/pull/505/commits/05cfe3b862d501b12528fc3119ee0e6a46139b4f).

takahirom (approved):
Thanks for your awesome investigation!

PR #534, opened 2022-09-17
Add a yml file for workflow_run trigger

### LINKED ISSUE #497: Separate CI workflow to post preview test comment
**Idea Description**
Now, if you don't fork repository and create a pull request, you can see the diff of a screenshot.
https://github.com/DroidKaigi/conference-app-2022/pull/471#issuecomment-1247638122
image
But we are using `pull_request` hook that uses the forked repository's TOKEN. So you can't post a comment to the pull request.
We should separate workflow to post a comment.
I think if we download the artifact and move comment to a new workflow, we can fix it.
https://github.com/DroidKaigi/conference-app-2022/blob/main/.github/workflows/ScreenShotTest.yml#L48-L55

There is a similar sample in the GitHub blog. Please see `CommentPR.yml`.
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

kwmt:
## Issue
- #497

## Overview (Required)
- I want to merge this PR in its current state. 
Because  I try to separate ScreenShotTest workflow and  I use  `workflow_run` trigger but I can’t check for the bellow reason.
After merge the PR, I’m going to do [the remaining work](https://github.com/DroidKaigi/conference-app-2022/pull/541).

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -1,0 +1,19 @@
+name: Comment on the pull request
+on:
+  workflow_run:
+    workflows:
+      - ScreenShotTest
+    types:
+      - completed
+
+jobs:
+  comment:
+    runs-on: ubuntu-latest
+    if: >
+      github.event.workflow_run.event == 'pull_request' &&
+      github.event.workflow_run.conclusion == 'success'
+    steps:
+      - name: Download artifact
+        run: echo "download artifact later"
+      - name: Comment PR
+        run: echo "comment pr later"
--- a/.github/workflows/ScreenShotTest.yml
+++ b/.github/workflows/ScreenShotTest.yml
@@ -1,3 +1,4 @@
+name: ScreenShotTest
 on:
   pull_request:
 


.github/workflows/CommentPR.yml
```diff
@@ -0,0 +1,19 @@
+name: Comment on the pull request
```
kwmt, line 1: I used [CommentPR.yml](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/) as a reference.

takahirom:
@kwmt Do you want to merge this first?

kwmt:
@takahirom Yes, that's right!

kwmt:
thanks!

PR #549, opened 2022-09-17
:wrench: Fix Modified bot filtering conditions

Corvus400:
## Issue
- Nothing.

## Overview (Required)
- ImgBotApp was included in the contributors, so we modified the filtering conditions so that it is not included.

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
bot can filter people.
https://www.worldsys.org/europe/search/?t%5B%5D=F&t%5B%5D=M&t%5B%5D=S&w=bot&match=contain&l%5B%5D=en&l%5B%5D=de&l%5B%5D=fr&l%5B%5D=it&l%5B%5D=es&l%5B%5D=sv&l%5B%5D=fi&l%5B%5D=ru&l%5B%5D=cs&l%5B%5D=nl&char_type=&char_length=&char_equal=&t%5B%5D=SF&limit=100&a=search
Probably we should filter it by the name one by one. Like "Imgbot" != name or !botSet.contains(name)

core-data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/ExcludedNames.kt
```diff
@@ -0,0 +1,9 @@
+package io.github.droidkaigi.confsched2022.data
+
+internal object ExcludedNames {
```
takahirom, line 3: I think you can use ExcludedGitHubUserNames. 👀

core-data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/ExcludedNames.kt
```diff
@@ -0,0 +1,9 @@
...
+    val LIST = listOf(
```
takahirom, line 4: I think there is not so much difference. But If you use setOf(), it is said that the time conplexity will O(1) instead of O(n) when you check contains() because Set uses hash under the hood.

takahirom (commented):
Almost LGTM! Thanks for addressing the issue!

core-data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/ExcludedNames.kt
Corvus400: @takahirom 
I see! That's interesting because I had no idea about the difference between setOf and listOf in terms of computational complexity!
I spent about 10 minutes searching the web for information describing it, but could not find it.
For the sake of posterity, if you don't mind, could you tell me where you got the quote from?

takahirom: I googled 👀 
image
https://www.bigocheatsheet.com/

And you can see the benchmark of it.
https://stackoverflow.com/a/10762995/4339442

takahirom (approved):
LGTM!

PR #569, opened 2022-09-18
Migrate project structure1: Migrate gradle structure

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- project(":feature-sessions") -> project("feature:sessions")

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


settings.gradle.kts
```diff
@@ -25,25 +25,32 @@ dependencyResolutionManagement {
...
+    "core-designsystem",
+    "core-data",
+    "core-testing",
+    "core-model",
+    "preview-screenshots"
 )
 
+modules.forEach { filePath ->
+    val (type, name) = filePath.split("-")
+    val newProjetName = ":$type:$name"
```
kittinunf, line 49: Using `:` might make the iOS bundle cringed ... iOS bundle use projectname so it would turn into something like `DroidKaigi2022.core:model.bundle` ... in my opinion using `.` is probably the best here.

takahirom: Thanks. I'm trying it 🙏 
https://github.com/DroidKaigi/conference-app-2022/pull/585

takahirom: For now, I decided to revert this.

takahirom: I managed to fix the issue through this PR. But your pr is still failing. 
https://github.com/DroidKaigi/conference-app-2022/pull/584/files
If we do this, It seems like our project can't use the project accessor.
https://github.com/DroidKaigi/conference-app-2022/pull/585

PR #587, opened 2022-09-18
hide the software keyboard when the drawer menu starts expanding.

### LINKED ISSUE #551: Hide the keyboard when the Drawer menu is expanding.
**Idea Description**
We can expand the Drawer menu even though it shows the software keyboard.
But I think this state is not good whether to input the keyboard or select a menu item from the Drawer menu.

So I suggest you hide the software keyboard when the Drawer menu opens.

nyafunta9858:
## Issue
- close #551 

## Overview (Required)
- hide the software keyboard when the drawer menu starts expanding.

## Links
- Nothing.

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
@@ -47,9 +47,11 @@ import androidx.compose.runtime.MutableState
 import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.ui.ExperimentalComposeUiApi
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.vector.ImageVector
 import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.platform.LocalSoftwareKeyboardController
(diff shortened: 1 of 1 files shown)


app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
```diff
@@ -203,8 +209,13 @@ fun KaigiAppDrawer(
             }
         }
     } else {
+        val keyboardController = LocalSoftwareKeyboardController.current
+        val drawerState = kaigiAppScaffoldState.drawerState
+        if (drawerState.isAnimationRunning && drawerState.isClosed) {
```
nyafunta9858, line 214: It may not need to be strict this far🤔 
(It may be enough to judge hiding the software keyboard to use only `isAnimationRunning.`)

takahirom: I thinks this can be called a lot. How about using LaunchedEffect?

nyafunta9858: Thank you for your advice!
I modified the following, could you please review it?
63310982d6e7f872b7c6c5fea91050bc474dca9d

takahirom (approved):
This implementation is useful! Thanks!

PR #611, opened 2022-09-19
:recycle: Update Added display of text when message is not null.

### LINKED ISSUE #601: Add session detail message
**Idea Description**
There can be a session that is canceled.
We need to inform to attendees.
So Please add the text under the title when there is a Session.message field.
We need this feature by 9/23.

image
https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=1977%3A10268

Corvus400:
## Issue
- close #601 

## Overview (Required)
- As the title says.

## Links
- https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=1977%3A2874

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt
@@ -59,6 +59,7 @@ import io.github.droidkaigi.confsched2022.designsystem.components.KaigiTag
 import io.github.droidkaigi.confsched2022.designsystem.theme.KaigiTheme
 import io.github.droidkaigi.confsched2022.designsystem.theme.TimetableItemColor
 import io.github.droidkaigi.confsched2022.model.Lang
+import io.github.droidkaigi.confsched2022.model.MultiLangText
 import io.github.droidkaigi.confsched2022.model.TimetableAsset
 import io.github.droidkaigi.confsched2022.model.TimetableItem
 import io.github.droidkaigi.confsched2022.model.TimetableItem.Session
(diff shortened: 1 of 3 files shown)


Corvus400:
I wanted to add a unit test, but it didn't work...🥺
スクリーンショット 2022-09-19 17 04 25
スクリーンショット 2022-09-19 17 04 39

takahirom:
Probably we should use createEmptyComposeRule() or own AndroidComposeTestRule() for it.

https://stackoverflow.com/questions/68267861/add-intent-extras-in-compose-ui-test
https://stackoverflow.com/a/69920214/4339442

Our purpose for the Robot testing pattern is to separate how and what. So if we can do it with the createEmptyComposeRule or something, it would be the best. 👀 

But you don't have to do this PR!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/SessionDetail.kt
```diff
@@ -348,33 +350,53 @@ fun SessionScheduleInfo(
...
-
-    val month = if (Locale.getDefault().language == japanese) {
+    val month = if (Locale.getDefault().language == Lang.JAPANESE.tagName.lowercase()) {
         "${sessionStartDateTime.monthNumber}月"
     } else {
         sessionStartDateTime.month.name.lowercase().replaceFirstChar { it.uppercase() }
     }
 
-    val day = if (Locale.getDefault().language == japanese) {
+    val day = if (Locale.getDefault().language == Lang.JAPANESE.tagName.lowercase()) {
```
takahirom, line 359: 👍

Issue #177, opened 2022-09-05
apply floor map  screen design


cOnigashima:
**Idea Description**
apply floor map  screen design

**Reference images and links**
coming soon

kubode:
✋ 


I have a few questions.

- Is this URL the correct screen design for the floor map?

    https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=1977%3A2763
    <img src="https://user-images.githubusercontent.com/5885032/190970416-7cffd6dc-583b-437a-8f78-799a3e97f2fb.png" width="240" />

- Does an API exist? Does it display an image of the asset? Or does it display an image of a specific URL?

- What happens when I tap this icon?

    ![image](https://user-images.githubusercontent.com/5885032/190971645-5348ee4a-b97d-4993-af38-6aefb029c2a2.png)

takahirom:
The floor map image is decided.
I didn't notice the pin icon. I'll talk with the designer about it👀 .

takahirom:
The pin icon is for zooming out to see the overview image. This image can be zoomed by pinching.
I think if you do this issue, you don't have to implement the icon first.

kubode:
You may ignore this.

After implementing the design, I thought a bit about this specification.

My opinion is that the pin icon is unnecessary if all we want to do is zoom in on an image.
This is because the feature to zoom in on an image is usually done by tapping on the image itself.
Also, this icon <img src="https://user-images.githubusercontent.com/5885032/190971645-5348ee4a-b97d-4993-af38-6aefb029c2a2.png" width="16" /> is usually used to display the current location.

From the above, I think it is better to perform the feature of zooming in on an image when the image is tapped, and to turn off the icon.

takahirom:
@kubode It makes sense, I'll talk with the designer about this. 🙏

PR #617, opened 2022-09-19
Implement the floor map screen

kubode:
## Issue
- Close https://github.com/DroidKaigi/conference-app-2022/issues/177

## Overview (Required)
- Implements the floor map screen.

### Design

[Figma](https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=1977%3A2763)

## TODO

> The floor map image is decided within 2, 3 days.

Using a dummy image and its URL on this PR.

> The pin icon is for zooming out to see the entire image. This image can be zoomed by pinching.
> I think if you do this issue, you don't have to implement the icon first.

I'll try this feature on the next PR.


takahirom:
Thanks! Could you check `./gradlew composeLint`?

kubode:
It's should be fixed.
Please check it again 🙏

feature/map/src/main/java/io/github/droidkaigi/confsched2022/feature/map/Map.kt
```diff
@@ -1,30 +1,112 @@
...
+        ) {
+            AsyncImage(
+                model = MapImageUrl,
+                modifier = Modifier
+                    .padding(16.dp)
+                    .fillMaxSize(),
+                contentScale = ContentScale.Fit,
+                alignment = Alignment.Center,
+                contentDescription = "Floor map",
+            )
```
takahirom, line 86: The release date is quite close and may be released as is. Can the icon be commented out for this reason?

takahirom (approved):
Looks great! Thanks!

feature/map/src/main/java/io/github/droidkaigi/confsched2022/feature/map/Map.kt
kubode: > Can the icon be commented out for this reason?

Do you mean remove the pin icon, which I pointed out in this comment?


I'm going to do this at https://github.com/DroidKaigi/conference-app-2022/issues/688

PR #634, opened 2022-09-20
[iOS] Remove redundant import

### LINKED ISSUE #589: [iOS] Remove redundant import in Model library
**Idea Description**

In Model library, there are many code`import appioscombined`, but it is not necessary because of `@_exported import appioscombined` written in `Model.swift` (see [here](https://github.com/DroidKaigi/conference-app-2022/blob/main/app-ios/Sources/Model/Model.swift)).

That's why I propose that removing all of `import appioscombined` code from Model library to make program simplify.

NakaokaRei:
## Issue
- close #589

## Overview (Required)
- removed redundant import in Model library


ry-itto (approved):
LGTM! Thanks 👍🏼

PR #641, opened 2022-09-21
Update: limit text size in timetable

### LINKED ISSUE #344: The session's titles become too small when they shrink.
**Idea Description**
We can define minimum text size. I think we can use 4 or 8 or 12 sp. Can you prototype and decide it?
I think we can calculate the size by (16sp / verticalScale) or 16sp * verticalScale and change the scale.
https://github.com/DroidKaigi/conference-app-2022/blob/f799ad158e73db9d28db579d2b4c7b9d489a6e92/feature-sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt#L41

[device-2022-09-10-203636.webm](https://user-images.githubusercontent.com/1386930/189481545-2b56528a-7501-4e27-8d09-9fa2f17d8111.webm)

warahiko:
## Issue
- close #344 

## Overview (Required)
- set a minimum value for the density of TimetableItem
  - calculated so that the text size of a title is at least 8 sp
  - if we use 12 sp, the layout of short items (less than 40 min) will be broken (in Pixel 6)
    - more layouts will be broken on smaller devices, but I think it would be difficult to deal with by changing the minimum text size

## Links
- 

## Screenshot
device: Pixel 6

Description | Before | After
:--: | :--: | :--:
font size: min |  | 
font size: max |  | 

ref. if we calculate so that the text size of a title is at least 12 sp
Description | Before | After (12 sp)
:--: | :--: | :--:
font size: min |  | 
font size: max |  |

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/TimetableItem.kt
@@ -42,8 +42,11 @@ fun TimetableItem(
     } else {
         Color(roomColor).copy(alpha = 0.2F)
     }
+    val titleTextStyle = MaterialTheme.typography.titleMedium
     val localDensity = LocalDensity.current.let {
-        Density(it.density * verticalScale, it.fontScale)
+        check(titleTextStyle.fontSize.isSp)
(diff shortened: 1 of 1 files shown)


takahirom:
Looks great! Thanks!

warahiko:
Thank you for reviewing!

PR #645, opened 2022-09-21
[iOS] [README] Update README for iOS project

kittinunf:
## Issue
- N/A

## Overview (Required)

This PR helps educate ppl to understand how to build the iOS project by updating app-ios/README.md

@ry-itto This is the file generated by the moko-resource. I don't see anything suspicious about the word "Android" except for the description of DroidKaigi itself (which is shoganai, in my opinion 😄). 

If you want to see this yourself, it is under XCFramework at

`build/XCFrameworks/debug/appioscombined.xcframework/ios-arm64/appioscombined.framework/DroidKaigi2022.core:model.bundle/Contents/Resources/Base.lproj/Localizable.strings`

Screen Shot 2022-09-21 at 19 18 21

Do you think we would have a problem with app store submission?

## Links
- N/A

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Thanks! This is so kind explanation about that!

> This is the file generated by the moko-resource. I don't see anything suspicious about the word "Android" except for the description of DroidKaigi itself (which is shoganai, in my opinion 😄).

Okey, in my experience, if there is the word "Android", it will be rejected, so I would like to at least remove the word "Android" from moko-resource.
However once I'll submit app to AppStoreConnect and wait for reviewing. If it rejected, I'll fix it with above way.

ry-itto:
@kittinunf 
Is this okey for merge? 👀

kittinunf:
Go go!

PR #651, opened 2022-09-21
[iOS] Retire Strings and finish the migration

kittinunf:
## Issue
- N/A

## Overview (Required)
As the move over the StringResource into the KMM module and now we have completed the migration for using String in moko-resource as a single source of truth. To complete the migration on iOS side, it is time to retire the Strings module

I also add the simple extension to make the StringResource usage on iOS to be a bit simpler to use.

## Links
- 

## Screenshot

In 日本語

In English


app-ios/Package.swift
```diff
@@ -190,20 +185,9 @@ var package = Package(
             name: "StaffFeature",
             dependencies: [
                 .target(name: "Model"),
-                .target(name: "Strings"),
                 .product(name: "ComposableArchitecture", package: "swift-composable-architecture"),
             ]
         ),
-        .target(
```
kittinunf, line 197: We no longer need this module target.

app-ios/Sources/Strings/Bundle.swift
```diff
@@ -1,28 +0,0 @@
-// Workaround: https://developer.apple.com/forums/thread/664295
```
kittinunf, line 1: I have removed the whole folder as we no longer need it, let me know what you think.

ry-itto: How about Xcode Preview?
This bundle is workaround for Xcode Previews crash.

ry-itto: Sorry for not noticing in the previous PR.
Talking about the scope of changes in the last PR, this bundle originally used in Strings was needed to avoid the crash.

ry-itto: Anyway, I noticed we should discuss in issues about this 🙏🏼  I'll move this topic to GitHub issue.
Sorry 🙇🏼

ry-itto (approved):
Looks great!! So clean 🗑️

PR #657, opened 2022-09-21
remove filter that is used to extract bot

RyuNen344:
## Issue
- close nothing

## Overview (Required)
- backend started to filter bot user
- so, client app doesn't need to filter bot user

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/contributors/DataContributorsRepository.kt
+++ b/core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/contributors/DataContributorsRepository.kt
@@ -1,6 +1,5 @@
 package io.github.droidkaigi.confsched2022.data.contributors
 
-import io.github.droidkaigi.confsched2022.data.ExcludedGitHubUserNames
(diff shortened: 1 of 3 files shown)


core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/contributors/DataContributorsRepository.kt
```diff
@@ -18,7 +17,6 @@ public class DataContributorsRepository(
             send(
                 contributorsApi
                     .contributors()
-                    .filterNot { ExcludedGitHubUserNames.SET.contains(it.username) }
```
takahirom, line 21: Can we remove ExcludedGitHubUserNames?

takahirom (approved):

core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/contributors/DataContributorsRepository.kt
RyuNen344: yes we can!

PR #668, opened 2022-09-22
:sparkles: Add Added processing when data acquisition fails on the notification screen.

Corvus400:
## Issue
- Nothing.

## Overview (Required)
- I tried to display SnackBar when State is Error.
- I added the process as a trial and would like to introduce it to other screens if the direction to display the SnackBar is acceptable.

## Links
- Nothing.

## Movie

https://user-images.githubusercontent.com/13657682/191670119-765f0ae8-17cc-4e20-8093-cd51e16fc8aa.mp4


feature/announcement/src/main/java/io/github/droidkaigi/confsched2022/feature/announcement/AnnouncementsViewModel.kt
```diff
@@ -32,4 +33,10 @@ class AnnouncementsViewModel @Inject constructor(
             AnnouncementsUiModel(data)
         }
     }
+
+    fun onRetryButtonClick() {
+        viewModelScope.launch {
+            // TODO: Re-acquire data.
```
Corvus400, line 39: We were unable to implement the process of reacquisition.🥺
I need help with this part.🙇‍♂️

feature/announcement/src/main/java/io/github/droidkaigi/confsched2022/feature/announcement/Announcements.kt
```diff
@@ -220,6 +257,19 @@ fun FullScreenLoading(
...
 
+@Composable
+fun SnackBarWithError(
+    snackBarData: SnackbarData
+) {
+
+    Snackbar(
+        containerColor = MaterialTheme.colorScheme.primaryContainer,
+        actionColor = MaterialTheme.colorScheme.primary,
+        contentColor = MaterialTheme.colorScheme.onPrimaryContainer,
```
Corvus400, line 268: Colors, etc. are tentative.
They will need to be changed when the design comes out.

takahirom:
Thanks for addressing this problem.
I had an idea about this, and I'm a little particular about this. That's why there is no implementation of this.

If the state has an Error, I think the UI should show the UI Model's data. So if it is the error UI Model, I think we should always show something like error module instead of Snackbar.
If you want to show Snackbar, I think we should toggle UI Model like this.
https://developer.android.com/topic/architecture/ui-layer/events#consuming-trigger-updates

takahirom:
How about adding `retrySuggestion:Boolean` field in ui model? And toggle like this 👀 
https://developer.android.com/topic/architecture/ui-layer/events#consuming-trigger-updates

Corvus400:
@takahirom 
Understood, I will try again tomorrow.
I am not sure I understand how you want to do one thing, so please let me confirm.
> So if it is the error UI Model, I think we should always show something like error module instead of Snackbar.
What kind of UI is this "error module"?
So far, I have not been able to imagine what it would look like since there is no design design design.
So I would like to know more details.

takahirom:
Because I also don't have an idea about  error module.😅

Corvus400:
@takahirom 
Sorry, this is not the fix you suggested, but I fixed it.
It appears to be working on the face of it...🤔
Could you take another look at it?🙏

output

takahirom:
Thanks! Your codes are the base of our error handling implementation! Without your effort, we would have the error handring mechanism.
We are working on it in this PR. It will merge soon. If you can apply it to other screens, can you support us?
https://github.com/DroidKaigi/conference-app-2022/pull/702

Corvus400:
@takahirom 
> We are working on it in this PR. It will merge soon. If you can apply it to other screens, can you support us?
Understood!🙆‍♂️
I will work on it after the #702 merge.

takahirom:
@Corvus400 I'll take a look at the sessions screen. 🙏

Corvus400:
@takahirom 
Am I correct that you want me to refer to #718 and #702 and incorporate them into other screens?

takahirom:
@Corvus400 Yes, that is exactly what I mean. But I can make issues so you don't have to do that if you don't want to.👍

Corvus400:
@takahirom 
We will start putting implementation in place today in the areas that are TODO at this point.

PR #680, opened 2022-09-22
Close bottom sheet when click search text field

### LINKED ISSUE #675: Please close bottom sheet when you click search text field
**Idea Description**
Currently, we can see the keyboard with the bottom sheet⌨️ 

image

nanaten:
## Issue
- close #675 

## Overview (Required)
- can closed bottom sheet when the SearchTextField clicked.
- I used interactionSource to achieve click detection. Because `clickable` doesn't work.

## Links
- https://stackoverflow.com/questions/67902919/jetpack-compose-textfield-clickable-does-not-work
- https://developer.android.com/reference/kotlin/androidx/compose/foundation/interaction/InteractionSource

### Before
https://user-images.githubusercontent.com/46647326/191802544-cfcd5204-0131-467b-ad7b-231c8711f43b.mp4

### After
https://user-images.githubusercontent.com/46647326/191802790-21c7c24c-c1d0-4350-9044-99e8331fc0f0.mp4

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
@@ -4,6 +4,8 @@ import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.horizontalScroll
+import androidx.compose.foundation.interaction.MutableInteractionSource
+import androidx.compose.foundation.interaction.PressInteraction
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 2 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
```diff
@@ -218,14 +225,24 @@ private fun SearchScreen(
...
     val keyboardController = LocalSoftwareKeyboardController.current
     val focusRequester = remember { FocusRequester() }
+    val interactionSource = remember { MutableInteractionSource() }
+
     LaunchedEffect(Unit) {
         focusRequester.requestFocus()
     }
+    LaunchedEffect(interactionSource) {
+        interactionSource.interactions.collect {
+            if (it is PressInteraction.Release) {
```
takahirom, line 241: Can I ask why you use PressInteraction instead of click? I think you have some reason. I'd like to know it for my information.

takahirom (approved):
Looks great! Thanks!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
nanaten: The reason is that we feel it is more natural to judge the completion of the click action when the finger is released.
There is no problem with the press interaction, so it can be replaced 👍

PR #683, opened 2022-09-22
add: Github Actions: Delete old branch saving screenshots

### LINKED ISSUE #650: Add GitHub Action to delete old `companion_` branch
**Idea Description**

The `companion_` branch is for saving Compose preview diff images.
But we have to remove the branches for repository size.
So I would like to delete the old branch. Say 2 weeks ago commit or older.
For DroidKaigi, it is not needed. But if we introduce this preview mechanism to the production application, this is needed.

higekick:
## Issue
- close #650 

## Overview (Required)
- Added delete script at the end of ScreenShotTest action.
- 2 weeks expired term can be changed at env parameter if some other term is preferred.
- I am not sure if it works correctly but I just confirmed on my own repository..!

Thank you for the review !

### DIFF
--- a/.github/workflows/ScreenShotTest.yml
+++ b/.github/workflows/ScreenShotTest.yml
@@ -94,3 +94,22 @@ jobs:
         with:
           message: ${{ steps.pr_comment.outputs.comment }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Delete old branch saving screenshots
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EXPIRED_TERM: 2w # first letter of year, month, week and day
+          REMOTE_REPO: origin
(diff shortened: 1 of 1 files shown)


.github/workflows/ScreenShotTest.yml
```diff
@@ -94,3 +94,22 @@ jobs:
         with:
           message: ${{ steps.pr_comment.outputs.comment }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Delete old branch saving screenshots
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          EXPIRED_TERM: 2w # first letter of year, month, week and day
+          REMOTE_REPO: origin
+          TARGET_BRANCH_PREFIX: companion_
```
takahirom, line 103: I didn't think of this approach. This is configurable!

PR #685, opened 2022-09-23
Improved clear input button touch target size

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Applied `IconButton` to improve target size and ripple area of input field clear button.

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
```diff
@@ -245,13 +246,14 @@ private fun SearchTextField(
...
         },
         trailingIcon = {
-            Icon(
-                painter = painterResource(id = R.drawable.ic_delete),
-                contentDescription = "search_word_delete_icon",
-                modifier = Modifier.clickable {
-                    onSearchWordChange("")
-                }
-            )
+            IconButton(
```
takahirom, line 249: This is what we have to do 👍

PR #686, opened 2022-09-23
Improved sessions TopAppBar accessibility

NUmeroAndDev:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- Improved read aloud in talkback.  

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/core/designsystem/build.gradle.kts
+++ b/core/designsystem/build.gradle.kts
@@ -11,6 +11,7 @@ kotlin {
 }
 
 dependencies {
+    implementation(projects.core.model)
     kotlinCompilerPluginClasspath(libs.composeCompiler)
     implementation(libs.coilCompose)
     implementation(libs.composeUiGoogleFonts)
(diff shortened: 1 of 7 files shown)


core/designsystem/build.gradle.kts
```diff
@@ -11,6 +11,7 @@ kotlin {
 }
 
 dependencies {
+    implementation(projects.core.model)
```
NUmeroAndDev, line 14: I want to access String resources in the designsystem, so I added a dependency, is there any problem?

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -377,7 +379,7 @@ fun SessionsTopBar(
                 Image(
                     modifier = Modifier.size(30.dp),
                     imageVector = ImageVector.vectorResource(id = CoreR.drawable.ic_app),
-                    contentDescription = "logo in toolbar"
+                    contentDescription = null
```
NUmeroAndDev, line 382: null is set because it is not necessary to focus and read.

takahirom:
Our tests depend on the info. Can you take a look at it?

NUmeroAndDev:
Fixed the test code. 🙏

takahirom (approved):
Looks great!

PR #699, opened 2022-09-23
Open floor map from session detail ▲

### LINKED ISSUE #485: Open floor map from session detail ▲
**Idea Description**
Now It makes the app crash.
image

takathemax:
## Issue
- close #485

## Overview (Required)
- When the ▲ on the session details screen is tapped, the transition to floor map screen is made.

## Links
- Nothing

## Screenshot
Before | After
:--: | :--:
 |


takahirom (approved):
This is what we wanted!

PR #702, opened 2022-09-23
Refactor announcement

takahirom:
## Issue
- close #ISSUE_NUMBER

## Overview (Required)
- 

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


core/data/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/data/announcement/DataAnnouncementsRepository.kt
```diff
@@ -2,22 +2,29 @@ package io.github.droidkaigi.confsched2022.data.announcement
...
-            awaitClose { }
+            announcementStateFlow.collect {
+                send(it)
+            }
         }
     }
+
+    override suspend fun refresh() {
+        announcementStateFlow.value = announcementsApi
+            .announcements(JAPANESE.name.lowercase())
```
takahirom, line 28: We should fix this. I will create another PR

PR #719, opened 2022-09-24
:recycle: Update Added behavior when a URL is included in a notification item.

Corvus400:
## Issue
- Nothing.

## Overview (Required)
- When an announcement contains a URL, the user is redirected to the URL page.

## Links
- Nothing.

## Movie

https://user-images.githubusercontent.com/13657682/192079076-49b7852b-561f-4a01-8452-104f34e4e07d.mp4

https://user-images.githubusercontent.com/13657682/192079085-e243158f-db30-4bd1-a50f-a0bee2e63c36.mp4


takahirom (approved):
Looks great! There is a tweet that if we have multiple URLs, it doesn't work. But I think it is better.

PR #733, opened 2022-09-24
[iOS] Update the map icon image

woxtu:
## Issue
- N/A

## Overview (Required)
- This PR updates the map icon image in the session detail view

## Links
- [App Design](https://www.figma.com/file/NcSMs6dMsD88d4wOY0g3rK/DroidKaigi-2022-Conference-App?node-id=1977%3A10268)

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (commented):
Thanks!!

But I think `app-ios/Sources/Assets/Resources/Assets.xcassets/map.imageset/map.pdf` this file is already added 👀 
Can you revert this change?

woxtu:
This change only updates the image and does not add new one.
The reason for the update is that the color of the image is different from the design.

ry-itto (approved):
I see! Okey!

LGTM 👍🏼

Issue #734, opened 2022-09-24
Fix deleting screenshot GitHub Action job


takahirom:
**Idea Description**

https://github.com/DroidKaigi/conference-app-2022/blob/main/.github/workflows/CommentPR.yml#L106

```
fatal: not a git repository (or any of the parent directories): .git
Error: Process completed with exit code 128.
```

**Reference images and links**
https://github.com/DroidKaigi/conference-app-2022/actions/runs/3117853645/jobs/5056772576

nashcft:
🙋

takahirom:
Thanks! Assigned 👍 You can check it by creating test pr.

nashcft:
I think #754 resolved this issue:

https://github.com/DroidKaigi/conference-app-2022/actions/runs/3122425253/jobs/5064355002

takahirom:
Looks great! Thanks!

PR #738, opened 2022-09-24
[test] fix action CommentPR

nashcft:
test for #734

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -102,6 +102,7 @@ jobs:
           TARGET_BRANCH_PREFIX: companion_
           DATE_FORMAT: '%Y%m%d'
         run: |
+          cd temp_docs/
           expired_date=`date --date="$EXPIRED_TERM" "+${DATE_FORMAT}"` 
           git fetch --all
           git branch -r | grep "[^* ]+" -Eo | grep "${REMOTE_REPO}/${TARGET_BRANCH_PREFIX}.+" -Eo |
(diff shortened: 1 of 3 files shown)


takahirom:
Probably you should add some changes for running the comment action.

takahirom:
Like this
https://github.com/DroidKaigi/conference-app-2022/pull/724/files

nashcft:
Thanks for the information 🙏

takahirom:
Sorry. Probably you need to merge for testing.

nashcft:
ah, sorry. I just doubted my understanding and confirmed what the comment job did.

nashcft:
so we'll revert this test PR after merge and test with the change, right?

takahirom:
@nashcft
We have to merge the workflow changes. But we don't need to merge design change to main.
The changes in pr are used and the workflow is used in main.
So you can open two PRs that change workflow to merge and change the design to test.

nashcft:
@takahirom 

Let me confirm my understanding of what to do 🙏 

First, we need to open following two PRs:

1. includes only the change to the workflow yml file `CommentPR.yml` to fix #734
2. includes some changes to UI layout to test if the fix above works as expected

Then what we'll do for the test are...

1. open and merge PR 1 to apply the fix into `main` branch
2. open PR 2 to test the merged change
3. if the result of test is...
  - successful: the issue has been solved 🎉 
  - failure: ok let's try another fix

please tell me if something in my understanding wrong 🙏

takahirom:
That's exactly what I mean!

nashcft:
Thank you! Then I'd like to rework this as new PRs along https://github.com/DroidKaigi/conference-app-2022/pull/738#issuecomment-1257157393

Issue #526, opened 2022-09-16
Opening and closing the session details screen repeatedly causes the app to crash with an OOM error.


Corvus400:
**Describe the bug**
I don't know how to deal with this yet without further investigation, but the titular phenomenon occurs.

**To Reproduce**
1. open the Session List screen
2. display the session details screen one by one and return to it
3. continue to do 2. although the operation becomes slower
4. finally, no more operations are accepted and OOM error occurs and the app crashes

**Expected behavior**
1. the screen does not get choppy when opening the session details screen
2. no matter how many times I open the session detail screen, the app does not crash with an OOM error

**Screenshots**

```
FATAL EXCEPTION: OkHttp Dispatcher
                                                                                                    Process: io.github.droidkaigi.confsched2022.dev, PID: 20572
                                                                                                    java.lang.OutOfMemoryError: Failed to allocate a 370792 byte allocation with 76488 free bytes and 74KB until OOM, target footprint 268435456, growth limit 268435456
                                                                                                    	at java.lang.StringFactory.newStringFromUtf8Bytes(Native Method)
                                                                                                    	at java.lang.StringFactory.newStringFromBytes(StringFactory.java:80)
                                                                                                    	at java.lang.StringFactory.newStringFromBytes(StringFactory.java:101)
                                                                                                    	at okio.Buffer.readString(Buffer.kt:313)
                                                                                                    	at okio.Buffer.readString(Buffer.kt:302)
                                                                                                    	at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.kt:270)
                                                                                                    	at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
                                                                                                    	at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
                                                                                                    	at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1137)
                                                                                                    	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:637)
                                                                                                    	at java.lang.Thread.run(Thread.java:1012)
2022-09-16 23:33:01.011 20572-20733 Process                 io.....droidkaigi.confsched2022.dev  I  Sending signal. PID: 20572 SIG: 9
[...]

PR #739, opened 2022-09-24
Using method reference for better memorization.

eneim:
## Issue
- Attempt to close 526, but I think this PR is not enough.

## Overview (Required)
- AS-IS: callback using `viewModel.method` are not effectively remembered, because `viewModel` is not a stable type.
- TO-BE: using method reference instead because method reference capture the viewModel reference when it (method referenc) is created, so the value of `viewModel` doesn't affect the stabability of method reference. In another word, it is stable.

## Links
- More details about investigating this can be found here: https://github.com/DroidKaigi/conference-app-2022/issues/526#issuecomment-1251030842

## Screenshot
Before | After
:--: | :--:
<video src="https://user-images.githubusercontent.com/1776230/192101786-0ce834ba-bca1-4839-a51f-603d50910aa4.mov">|<video src="https://user-images.githubusercontent.com/1776230/192101815-628da50c-18a2-483f-b9ad-3dfe3ff9616b.mov">

### DIFF
--- a/app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
+++ b/app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
@@ -107,9 +107,7 @@ fun KaigiApp(
                 drawerSheetContent = {
                     DrawerSheetContent(
                         selectedDrawerItem = kaigiAppScaffoldState.selectedDrawerItem,
-                        onClickDrawerItem = { drawerItem ->
-                            kaigiAppScaffoldState.navigate(drawerItem)
-                        }
+                        onClickDrawerItem = kaigiAppScaffoldState::navigate,
                     )
                 }
             ) {


app-android/src/main/java/io/github/droidkaigi/confsched2022/KaigiApp.kt
```diff
@@ -107,9 +107,7 @@ fun KaigiApp(
                 drawerSheetContent = {
                     DrawerSheetContent(
                         selectedDrawerItem = kaigiAppScaffoldState.selectedDrawerItem,
-                        onClickDrawerItem = { drawerItem ->
-                            kaigiAppScaffoldState.navigate(drawerItem)
-                        }
+                        onClickDrawerItem = kaigiAppScaffoldState::navigate,
```
takahirom, line 110: May I ask if we can fix the problem by adding `Immutable` to KaigiAppScaffoldState? 👀

eneim: Good idea. Since `KaigiAppScaffoldState` as a `DrawerState` which is `Stable`, should we use `Stable` for the `KaigiAppScaffoldState` instead of `Immutable`?

takahirom: I think so too!
But I think we should have both of codes, method reference and stable. So now, I merge this PR. Thanks

eneim: > we should have both of codes, method reference and stable

@takahirom Agree. Thanks for merging. If you need any follow-up please assign me to a ticket.

PR #747, opened 2022-09-25
[iOS] [Android] Add moko resource to make color consistent accoss iOS / Android

kittinunf:
## Issue
- N/A, this is a foundation work.

## Overview (Required)
- This is a continuation over #387! Now, we are able to share color for both iOS and Android.

Color is defined in one place here.

Screen Shot 2022-09-25 at 12 57 33

If this is something we want to pursue, I will migrate color usage on iOS side and deprecated [Theme]

### Usage of Color on iOS side

```Swift

ColorsKt.shared.<color_name>.color()

```

### On Android side 

we should prefer getting color from MaterialTheme as much as possible. In this PR, I also fix some of the inconsistency inside the usage;

```Kotlin
MaterialTheme.colorScheme.<color_name>
```

This makes the color name on both iOS & Android consistent as we will refer to it from the figma by the designer; So the name is consistent across the board.

https://www.figma.com/file/XsVzpDZSkEQANgCZLRTDDy/DroidKaigi-2022-Material-3-Design-Kit?node-id=51510%3A8890

Screen Shot 2022-09-25 at 12 55 30

## Links
- 

## Screenshot

iOS

There should be no changes on the iOS side but now the color is coming from the moko-resources same source as in Android side.


app-android/src/main/res/values/colors.xml
```diff
@@ -1,12 +0,0 @@
-<?xml version="1.0" encoding="utf-8"?>
```
kittinunf, line 1: We deprecate this in favor of color in MR (inside core/model)

takahirom:
I would like to share about colors.
In this project, we would like to respect Android, and iOS design guidelines.
So currently we are sharing only Color tokens. So we don't want to share the theme.
What do you think?
https://github.com/DroidKaigi/conference-app-2022/blob/main/core/designsystem/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/designsystem/theme/KaigiColors.kt

kittinunf:
Understood. Just to be clear, the color that we share is Droidkaigi theme color set right? As you can see them in the "Theme" package in the iOS app. 

I think the Droidkaigi should be shared in iOS and Android correct?

In this PR, we are sharing them with moko resources. Do you think it's good idea?

takahirom:
I'll talk with @ry-itto 🙏

takahirom:
I talked with @ry-itto and @NUmeroAndDev .
In ios app, it is difficult to apply the elevation color which is implemented by mixing color, so the iOS app doesn't use the KaigiColor and use Eyedropper tool to find the color.
So currently our approach is to use KaigiColor as much as we can and if we can't do, we use platform defined colors.

kittinunf:
Ok then. no worries, it looks like we might not share the color resources this way. I will close this PR then.

takahirom:
Thanks for your great challenge!

kittinunf:
@takahirom what do you think that we should do next for moko-resources? I think if we don't share color (while it is a bit of a bummer), we can share image font/resources. Do you think it is worthwhile to explore on those?

takahirom:
@ry-itto What do you think? I think images would be good.

ry-itto:
@takahirom 

> @ry-itto What do you think? I think images would be good.

Sorry for late reply 🙏🏼 
I think fonts are good. But I think images are not good. I have concerns about this.

Almost all icons in this App are material icon. iOS project adds those icon files into resources, but I think Android project isn't.
Therefore, even if images are shared with moko-resources, iOS project need to add icons. This is duplicate definitions about image resource.

So I think we should explore sharing font resources with moko-resources.

kittinunf:
👌 @ry-itto and @takahirom thank you both for your valuable input. I will continue exploring with on image whether we could share or font then.

Since the color is not useful for sharing, I am closing this PR.

PR #749, opened 2022-09-25
Add current time line on timetable

### LINKED ISSUE #711: Add current time line on Timetable
**Idea Description**
It must be fun 👍 

image

iiinaiii:
## Issue
- close #711

## Overview (Required)
- Add current time line on Timetable
- Draw line and circle on Timetable Composable, and draw circle on Hours Composable.
- I used drawWithContent to draw a line in front of Content. (If drawing back is the specification, I would do it in drawBehind, but would drawWithContent be OK?)
- The current time line is calculated and drawn using only time and does not take the day into account, is that OK?
(The line is drawn no matter which day the app is opened.)
- The function to automatically update the line position while the screen is open is not implemented.

## Screenshot
Before | After
:--: | :--:
screen-shot-old | screen-shot-new

 |


takahirom:
Thank you for tacking this difficult issue.

> The current time line is calculated and drawn using only time and does not take the day into account, is that OK?
(The line is drawn no matter which day the app is opened.)

That could be confusing. Can you fix it? I'd like to do it like Google calendar.
And I'm a little worried about the performance. The time will change every compose and the Instant is unstable type so you can't skip recompose.
Something like this.

```kotlin
data class SessionsUiModel (
...
val timeLine: TimeLine
)

@Immutable
data class TimeLine(val instant: Instant, val day: DroidKaigiDay)
```

iiinaiii:
Thanks for the comment.

> That could be confusing. Can you fix it? I'd like to do it like Google calendar.
Understood! Please wait a moment while we try and see.

Thanks for the advice about performance too. I will try to fix it.

iiinaiii:
@takahirom 
>The current time line is calculated and drawn using only time and does not take the day into account, is that OK?
(The line is drawn no matter which day the app is opened.)

>That could be confusing. Can you fix it? I'd like to do it like Google calendar.

I fixed to take into account the day of the month.
The current time line is only displayed on the applicable day.

Applicable Date | Not Applicable Date
:--: | :--:
Screenshot_1006_1430 | Screenshot_1005_1430

iiinaiii:
And I have modified it to use immutable data class as you advised.

```kotlin
@Immutable
public data class TimeLine(
    private val instant: Instant,
    private val day: DroidKaigi2022Day,
) 
```

I also added the following code because I think it would be better if the current time line is updated every time the screen is opened.
Please point out any problems.

```kotlin
fun SessionsScreenRoot(
   ...
) {
   ...
    val lifecycleObserver = remember(viewModel) {
        LifecycleEventObserver { _, event ->
            if (event == Lifecycle.Event.ON_RESUME) {
                viewModel.onUpdateTimeLine()
            }
        }
    }
    val lifecycle = LocalLifecycleOwner.current.lifecycle
    DisposableEffect(lifecycle) {
        lifecycle.addObserver(lifecycleObserver)
        onDispose {
            lifecycle.removeObserver(lifecycleObserver)
        }
    }
```

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -92,17 +95,30 @@ fun SessionsScreenRoot(
...
     onNavigationIconClick: () -> Unit = {},
     onSearchClicked: () -> Unit = {},
     onTimetableClick: (TimetableItemId) -> Unit = {},
 ) {
     val state: SessionsUiModel by viewModel.uiModel
 
+    val lifecycleObserver = remember(viewModel) {
+        LifecycleEventObserver { _, event ->
+            if (event == Lifecycle.Event.ON_RESUME) {
+                viewModel.onUpdateTimeLine()
```
takahirom, line 107: Looks great. But I think we can make the name `onLifecycleResume`. Because it is the event of view.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
```diff
@@ -130,6 +150,7 @@ fun Sessions(
 ) {
     val scheduleState = uiModel.state
     val isTimetable = uiModel.isTimetable
+    val timeLineState = rememberTimeLineState(uiModel.timeLine)
```
takahirom, line 153: I think this is the duplicate single source of truth. I thought we don't have to have the state in Compose. Is there any problem about this? 👀

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
iiinaiii: That sounds better!
Fixed to onLifecycleResume.

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Sessions.kt
iiinaiii: Fixed without using TimeLineState within Compose 🙇 

I was inexperienced about State of Compose.
So, I had put the TimeLine in the val property and was unable to update it, so I had to use State.
I was able to do it by using uiModel.timeLine directly.

* NG
```kotlin
@Composable
fun Sessions(
    uiModel: SessionsUiModel,
...
){
     val timeLine = uiModel.timeLine
     ...
    Timetable(
        timeLine = timeLine
    )
}
```

* OK
```kotlin
@Composable
fun Sessions(
    uiModel: SessionsUiModel,
...
){
     ...
    Timetable(
        timeLine = uiModel.timeLine
    )
}
```

Thanks for the advice!

takahirom (approved):
Looks great! Thanks for your contribution! Everybody at the conference will use this!

PR #750, opened 2022-09-25
Ignore case for search

### LINKED ISSUE #676: Ignore case for search
**Idea Description**
If we want to search Compose. We can't search `compose`. We should search `compose`.
Can we make it ignore case?

image

CookieySun:
## Issue
- close #676

## Overview (Required)
- Ignore case for search
- Ignore case for Highlight

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |

### DIFF
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
@@ -62,8 +62,9 @@ public data class Timetable(
         }
         if (filters.searchWord.isNotBlank()) {
             timetableItems = timetableItems.filter { timetableItem ->
-                timetableItem.title.currentLangTitle.lowercase().contains(
-                    filters.searchWord.lowercase()
+                timetableItem.title.currentLangTitle.contains(
+                    filters.searchWord,
+                    ignoreCase = true
                 )
             }
         }
--- a/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
+++ b/core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
@@ -62,7 +62,9 @@ public data class Timetable(
         }
         if (filters.searchWord.isNotBlank()) {
             timetableItems = timetableItems.filter { timetableItem ->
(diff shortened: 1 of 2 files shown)


core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
```diff
@@ -62,7 +62,7 @@ public data class Timetable(
         }
         if (filters.searchWord.isNotBlank()) {
             timetableItems = timetableItems.filter { timetableItem ->
-                timetableItem.title.currentLangTitle.contains(filters.searchWord)
+                timetableItem.title.currentLangTitle.lowercase().contains(filters.searchWord.lowercase())
```
takahirom, line 65: Can you check it?

```
Step 'ktlint' found problem in 'src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt':

Error on line: 65, column: 1
Exceeded max line length (100)
java.lang.AssertionError: Error on line: 65, column: 1
Exceeded max line length (100)
```

CookieySun: I fix it:pray:

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
```diff
@@ -62,7 +62,9 @@ public data class Timetable(
         }
         if (filters.searchWord.isNotBlank()) {
             timetableItems = timetableItems.filter { timetableItem ->
-                timetableItem.title.currentLangTitle.contains(filters.searchWord)
+                timetableItem.title.currentLangTitle.lowercase().contains(
```
takahirom, line 65: How about using contains(text, ignoreCase = false)?

takahirom:
Looks good! Please take a look at small point🙏

core/model/src/commonMain/kotlin/io/github/droidkaigi/confsched2022/model/Timetable.kt
CookieySun: I don't know this option.
But, false is case sensitive. so I use true.

takahirom (approved):
Thanks!

PR #751, opened 2022-09-25
[iOS] Enable links in announcements

woxtu:
## Issue
- N/A

## Overview (Required)
- This PR adds a trick to enable links in announcements

## Links
- https://stackoverflow.com/questions/57744392/

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
I don't know about this!
This is so good! Thanks!

PR #754, opened 2022-09-25
change directory before running git command

nashcft:
## Issue
- #734
  - stay the issue open because we have to test if the change resolve it after this PR is merged

## Overview (Required)

The cause of error reported in #734 is running `git` command out of git repository. This occurred because current directory  when the step `Delete old branch saving screenshots` is executed is `$GITHUB_WORKSPACE` although the checked-out repository is located on `$GITHUB_WORKSPACE/temp_docs`.
The step worked fine before workflow was split in #723 because the base workflow (`ScreenShotTest`) also checked-out repository on `$GITHUB_WORKSPACE` to run screenshot tests.

This PR added directory change at the beginning of commands executed in `Delete old branch saving screenshots`. The PR also tried to make destinations of directory-change operations clear with the default environment variable `$GITHUB_WORKSPACE`.

## Links

N/A

## Screenshot

N/A

### DIFF
--- a/.github/workflows/CommentPR.yml
+++ b/.github/workflows/CommentPR.yml
@@ -50,7 +50,7 @@ jobs:
         env:
           BRANCH_NAME: companion_${{ github.event.workflow_run.head_branch }}
         run: |
-          cd temp_docs/
+          cd $GITHUB_WORKSPACE/temp_docs/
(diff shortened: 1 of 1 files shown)


takahirom:
Great!

nashcft:
now I'll open a test pull request 🏃

PR #788, opened 2022-09-28
mod data class property to fix serialization

RyuNen344:
## Issue
- close nothins

## Overview (Required)
- fix serialization model
- Speaker bio and tagLine need to be nullable

```log
Caused by: kotlinx.serialization.json.internal.JsonDecodingException: Unexpected JSON token at offset 176914: Unexpected 'null' value instead of string literal at path: $.speakers[28].bio
JSON input: .....e":"","lastName":"","bio":null,"tagLine":null,"profilePictur.....
```

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


takahirom (commented):
Ideally, I think we should make it nullable and delete the UI if it is optional. But, for now it is ok.🙆

takahirom (approved):

RyuNen344:
> Ideally, I think we should make it nullable and delete the UI if it is optional. But, for now it is ok.🙆

maybe, we are not using `bio` and `tagLine` now? 🤔

PR #799, opened 2022-09-29
:recycle: Update Enable transition from session details to the search screen.

Corvus400:
## Issue
- Nothing.

## Overview (Required)
- Tapping a category tag now takes you to the search screen.
- When the search screen is displayed, the search will already be performed using the tag displayed in the session details.

## Links
- Nothing.

## Movie

https://user-images.githubusercontent.com/13657682/192979241-9890e0b2-9933-4aa0-9472-cb2640dcbab7.mp4

https://user-images.githubusercontent.com/13657682/192979266-229d4727-a125-46d1-9de5-76bfe4280897.mp4

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
@@ -89,7 +89,7 @@ import kotlinx.datetime.toLocalDateTime
 @Composable
 fun SearchRoot(
     modifier: Modifier = Modifier,
-    categoryId: String? = null,
+    initialCategoryId: String? = null,
     viewModel: SearchViewModel = hiltViewModel(),
     onBackIconClick: () -> Unit = {},
     onItemClick: (TimetableItemId) -> Unit,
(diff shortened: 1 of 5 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
```diff
@@ -99,6 +100,10 @@ fun SearchRoot(
 
     val keyboardController = LocalSoftwareKeyboardController.current
 
+    if (categoryId != null) {
+        viewModel.onCategorySelected(categoryId)
```
takahirom, line 104: I think we shouldn't call viewModel moethod in Composable function directly. 👀 
So how about using LaunchedEffect?

LaunchedEffect(initialCategoryId) {
  viewModel.onCategorySelected(categoryId)
}

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Search.kt
```diff
@@ -100,8 +100,10 @@ fun SearchRoot(
 
     val keyboardController = LocalSoftwareKeyboardController.current
 
-    if (categoryId != null) {
-        viewModel.onCategorySelected(categoryId)
+    LaunchedEffect(categoryId) {
+        if (categoryId != null) {
+            viewModel.onCategorySelected(categoryId)
```
takahirom, line 105: Sorry, can we use initialCategoryId instead of categoryId? Because it is used only for the initial state.

takahirom:
Thanks! LGTM!

PR #801, opened 2022-09-29
[iOS] Fix consistency on String's localized() usage

kittinunf:
## Issue
- N/A. Fix bug

## Overview (Required)

We should prefer using `localized()` because of 2 reasons;

1. For consistency
2. `localized` also removes Android words

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Looks good 👍🏼 Thanks!

PR #804, opened 2022-09-29
[iOS] Enable case-insensitive searches

woxtu:
## Issue
- N/A

## Overview (Required)
- The title says it all

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
Thanks!! This is so good 👍🏼

PR #820, opened 2022-09-30
Change announcement title from News into Announcement

### LINKED ISSUE #803: Difference of words Annoucement on drawer and the screen's title, it is News
**Describe the bug**
In English setting, there is a difference of item name Annoucement on drawer and the screen's title, News.

**To Reproduce**
1. Set in English from drawer menu Setting.
2. Show drawer and see the item `Announcement`.
3. Tap the `Announcement`, and see the showed screen's title.

**Expected behavior**
Decide which is a better word, and set it.

**Screenshots**

ohzono:
## Issue
- close #803 

## Overview (Required)
- Change announcement title from News into Announcement.

## Links
- 

## Screenshot
Before | After
:--: | :--:
|

### DIFF
--- a/core/model/src/commonMain/resources/MR/en/strings.xml
+++ b/core/model/src/commonMain/resources/MR/en/strings.xml
@@ -23,7 +23,7 @@
     <string name="about_app_version">App Version</string>
 
     <!-- Announcement -->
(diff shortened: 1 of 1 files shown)


takahirom (approved):
Thanks for your proposal🙏

PR #822, opened 2022-09-30
Fix splash screen duration

gen0083:
## Overview (Required)

Splash animation looks like 3000ms but windowSplashScreenAnimationDuration 1000.

## Screenshot
Before | After
:--: | :--:
before | after

It looks no difference in debug mode...


takahirom:
Looks great!

PR #826, opened 2022-10-01
Fix session is not shown when session's time is longer than screen size

### LINKED ISSUE #824: If session start time or end time is longer than screen size session not shown
**Describe the bug**
If session start time or end time is longer than screen size session not shown.

**To Reproduce**
1. Go to session page on day3
2. Scroll down to 14:00
3. Scheduled session is not shown

**Expected behavior**
It it better to show that there is a session.

**Screenshots**
device-2022-10-01-160123 2022-10-01 16_09_05

gen0083:
## Issue
- close #824 

## Overview (Required)
- If session's time is longer than screen displayed time then session is not shown.

## Screenshot
Before | After
:--: | :--:
device-2022-10-01-160123 2022-10-01 16_09_05 | after


takahirom (approved):
Looks great!

PR #838, opened 2022-10-03
Revert "update: using downloadablefont instead of local file"

### LINKED ISSUE #802: Unnecessary ellipsis appear on the time-line title
**Describe the bug**
Time-line title ellipsis is displayed on first line, but it seems to be better to be displayed at maximum line only.

**To Reproduce**
1. Install from Play Store.  I haven't confirmed yet in development environment.
2. For example, set in Japanese from drawer menu `Setting`.
3. Go to Day 2 time line.
4. See `App bars`'s column at 12:00. There is a session `通知チャンネルに対応する`. 
5. See the behavior of the ellipsis like `通知チャンネルに対応...`.

**Expected behavior**
I think it would be better to show in 2nd line in the above case.
So, I suggest that the ellipsis is appeared just on maximum line.

**Screenshots**

Luecy1:
## Issue
- close #802

## Overview (Required)
- Two lines may not be displayed due to downloadable font
- Reverted to use local files instead of downloadable fonts

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
Please run `spotlessApply` 🙏 
And there are preview changes. This is because downloadable fonts are not applied when previewing, right?

Luecy1:
run & push deone.

About the preview
I'm not sure downloadable fonts are applied in the preview.
I think it was probably applied because I made it use a local font.

takahirom:
Thanks for addressing the problem!

Issue #795, opened 2022-09-29
Fix crash when calculating scroll


takahirom:
**Idea Description**

```
Fatal Exception: java.lang.IllegalStateException
Offset is unspecified
androidx.compose.ui.geometry.Offset.getX-impl (Offset.java:28)
io.github.droidkaigi.confsched2022.feature.sessions.ScreenScrollState$scroll$2$1.invokeSuspend (ScreenScrollState.java:31)
kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (BaseContinuationImpl.java:8)
kotlinx.coroutines.DispatchedTask.run (DispatchedTask.java:100)
```

<img width="867" alt="image" src="https://user-images.githubusercontent.com/1386930/192918412-07c267ab-fc41-4d7d-b83a-4b5314cd5d67.png">

```kotlin
    suspend fun scroll(
        dragAmount: Offset,
        timeMillis: Long,
        position: Offset,
    ) {
        val nextPossibleX = calculatePossibleScrollX(dragAmount.x)
        val nextPossibleY = calculatePossibleScrollY(dragAmount.y)
        scrollState.scroll(
            Offset(nextPossibleX, nextPossibleY), // ←
```

```kotlin
value class Offset internal constructor(internal val packedValue: Long) {

    @Stable
    val x: Float
        get() {
            // Explicitly compare against packed values to avoid auto-boxing of Size.Unspecified
            check(this.packedValue != Unspecified.packedValue) {
                "Offset is unspecified" // ←
            }
            return unpackFloat1(packedValue)
        }
```

**Reference images and links**

Tomoyuki-TAKEZAKI:
I'd like to work on this issue 🙋🏻

takahirom:
@Tomoyuki-TAKEZAKI I'm looking forward to seeing your PR! Assigned 👍

takahirom:
@Tomoyuki-TAKEZAKI Let me unassign issue 🙏

takahirom:
<img width="1345" alt="image" src="https://user-images.githubusercontent.com/1386930/194450448-130db35c-f037-4b67-87d4-cefbb156cecb.png">

Tomoyuki-TAKEZAKI:
@takahirom 
I was using Flamingo instead of Android Studio Electric Eel, so the build didn't go well and I was late.
I was going to tackle it during the third-day session of DroidKaigi.
Sorry :sad:

[...]

PR #878, opened 2022-10-07
[Bugfix] Check that scroll offset is not Float.NAN before scroll

Tomoyuki-TAKEZAKI:
## Issue
- https://github.com/DroidKaigi/conference-app-2022/issues/795

## Overview (Required)

The crash is Caused by creating an instance of Offset with `Float.NAN`, but I can't find the root cause of it.
In other words, I can't find a division by `0f` on the call hierarchy.

So I added an `if` block to prevent scroll with an invalid offset.

This does not fix the root cause, but I think this will work.

## Screenshot

I can't reproduce the crash on my env, so no screen recording.

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Hours.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Hours.kt
@@ -269,11 +269,13 @@ private class HoursScreen(
         position: Offset,
     ) {
         val nextPossibleY = calculatePossibleScrollY(dragAmount.y)
-        scrollState.scroll(
-            Offset(scrollState.scrollX, nextPossibleY),
-            timeMillis,
-            position
-        )
+        if (scrollState.scrollX.isNaN().not() && nextPossibleY.isNaN().not()) {
+            scrollState.scroll(
+                Offset(scrollState.scrollX, nextPossibleY),
+                timeMillis,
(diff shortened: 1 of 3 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Hours.kt
```diff
@@ -269,11 +269,13 @@ private class HoursScreen(
         position: Offset,
     ) {
         val nextPossibleY = calculatePossibleScrollY(dragAmount.y)
-        scrollState.scroll(
-            Offset(scrollState.scrollX, nextPossibleY),
-            timeMillis,
-            position
-        )
+        if (scrollState.scrollX.isNaN().not() && nextPossibleY.isNaN().not()) {
```
takahirom, line 272: Looks good! If we have time we can create `ScrollState.safeScroll()` extension function.

takahirom (approved):
Thanks!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Hours.kt
Tomoyuki-TAKEZAKI: I couldn't react to your comment immediately, and I couldn't reflect this comment.
Thank you for your proposal.

PR #888, opened 2022-10-07
[Detekt] Add ignore config for compose function

fumiya-kume:
## Issue
- none

## Overview (Required)

 Add detekt ignore config for Composable method.(method name should be Camel case)

## Links
- https://github.com/detekt/detekt/issues/2733

## Screenshot
Before | After
:--: | :--:
 |


takahirom:
I didn't know about "DeteKt 1.22.0-RC1". thanks!

PR #889, opened 2022-10-07
Add test cases for AboutScreenRobot to assert AuxiliaryInformation

rkowase:
## Issue
- n/a

## Overview (Required)
In this PR, I added test cases for AboutScreenRobot to make sure that the below elements are visible.
* Access
* Staff List
* License
* Code Conduct
* Privacy Policy
* App Version

## Links
- n/a

## Screenshot
Before | After
:--: | :--:
 | No changes (same as before)

The test has been passed locally. ✅ 
スクリーンショット 2022-10-07 16 25 12


takahirom (approved):
👍 
I think we don't have to check everything. But I think these features are important. So I think it is good!

PR #891, opened 2022-10-07
[iOS] Lint only the iOS app directory

woxtu:
## Issue
- N/A

## Overview (Required)
- This PR changes to run the linting action only when the iOS app directory has changes

## Links
- 

## Screenshot
Before | After
:--: | :--:
 |


ry-itto (approved):
thanks!👍

PR #895, opened 2022-10-09
adjustment for zipline v0.9.5

ked4ma:
## Issue
- nothing
- but related to #864 

## Overview (Required)
- zipline is renamed `LoadedZipline` to `LoadResult` as return value of `load` func.
- Fitting code to load timeline.

## Links
- https://github.com/cashapp/zipline/blob/0.9.5/zipline-loader/src/commonMain/kotlin/app/cash/zipline/loader/LoadResult.kt

## Screenshot
UI codes are not changed.


takahirom (approved):
Thanks for migration!

PR #898, opened 2022-10-10
[Improvement] introduce ScrollState#safeScroll

Tomoyuki-TAKEZAKI:
## Issue
- https://github.com/DroidKaigi/conference-app-2022/issues/795

## Overview (Required)

I introduced a method ScrollState#safeScroll which is suggested in the comment https://github.com/DroidKaigi/conference-app-2022/pull/878/files#r990602724

## Screenshot

No changed in behavior, so no screenshot.

### DIFF
--- a/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
+++ b/feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
@@ -192,13 +192,12 @@ fun Timetable(
                 scrollBy(
                     action = { x: Float, y: Float ->
                         coroutineScope.launch {
-                            if (x.isNaN().not() && y.isNaN().not()) {
-                                scrollState.scroll(
-                                    amount = Offset(x, y),
-                                    timeMillis = 0,
-                                    position = Offset.Zero
-                                )
-                            }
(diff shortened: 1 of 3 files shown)


feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
```diff
@@ -418,6 +417,21 @@ class ScreenScrollState(
         }
     }
 
+    suspend fun safeScroll(
```
takahirom, line 420: I didn't think it was our class. I think we can delete the `safeScroll` and use the implementation as the default implementation of the `scroll` method. Because it is scroll, so I think it can not be the critical feature.
What do you think?

But I think it is better than the current implementation. So please let me merge this for now!

takahirom:
Thank you for your cooperation!

feature/sessions/src/main/java/io/github/droidkaigi/confsched2022/feature/sessions/Timetable.kt
Tomoyuki-TAKEZAKI: Thank you for your comment!
I've opened a PR for the rename and simplification https://github.com/DroidKaigi/conference-app-2022/pull/902

PR #916, opened 2022-10-16
Update: reduce recomposition of TimeLane

### LINKED ISSUE #867: Performance improvement for Tabletable TimeLine
**Idea Description**
Now, our Tabletable TimeLine is recomposed by every scroll.
So we can probably use `drawBehind{}`?

image

warahiko:
## Issue
- close #867 

## Overview (Required)
TimeLane composable is recomposed by every scroll, so we should reduce recomposition.
- change `List` to `ImmutableList`
- `LazyListItemInfo#offset` is changed by every scroll, but it is used only for changing offset of Box composable, so we can pass it by lambda function
  - TimeLane composable will be recomposed by changing visible items
  - `LazyListItemInfo#index` is not changed by every scroll

## Links
- 

## Screenshot
No change in appearance

## Note

I have no idea how we can use `drawBehind()` to solve this problem.
If you have it, please give me advice.

I'm sorry for being late 🙏


takahirom:
👀

takahirom (approved):
Looks great!

